PR #33, opened 2014-10-22
Remove xen-netback/front modules for aarch64

dirkmueller:
They don't exist there as Xen paravirt support is disabled.


mvidner:
Is this still relevant?

wfeldt:
The patch is a no-op; this pr can be closed.

PR #42, opened 2015-02-03
Test zen1

wfeldt:
zenworks image adjustments

### DIFF
--- a/data/root/zenroot.file_list
+++ b/data/root/zenroot.file_list
@@ -51,6 +51,7 @@ else
   openSUSE-release-mini:
 endif
 
+?grub:
 bzip2:
 coreutils:
 cpio:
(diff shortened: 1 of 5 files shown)


data/root/zenroot.file_list
```diff
@@ -51,6 +51,7 @@ else
   openSUSE-release-mini:
 endif
 
+?grub:
```
jreidinger: why grub1?

jreidinger:
otherwise it looks good. Just grub1 do not make sense for me

data/root/zenroot.file_list
wfeldt: they still use it

jreidinger: OK, I hope they do not expect that we will support it.

jreidinger:
LGTM

PR #43, opened 2015-02-18
Add new kernel module ipvlan for kernel 3.19.0

DimStar77:
This should fix the build issue of installation-images currently seen at
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:D/installation-images-openSUSE/standard/x86_64 due to kernel 3.19.0


coolo:
steffen is on vacation, so in case this is not the right thing to do: please create a bug report for this issue so he reviews

DimStar77:
done: https://bugzilla.opensuse.org/show_bug.cgi?id=918385

PR #50, opened 2015-03-24
Fix dangling symlinks for ca-certificates (with p11-kit 0.23.x)

DimStar77:
It's supposed to fix
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:D/installation-images-openSUSE/standard/i586

and is hopefully the right way to do so.

### DIFF
--- a/data/base/base.file_list
+++ b/data/base/base.file_list
@@ -108,6 +108,8 @@ ca-certificates:
   /usr/lib/ca-certificates/update.d/*openssl.run
   /usr/lib/ca-certificates/update.d/*etc_ssl.run
   /var/lib/ca-certificates/openssl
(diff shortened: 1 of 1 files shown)


DimStar77:
Done (without a clue what this means and I also did not build test this locally)

DimStar77:
we were actually also wondering what might have changed for this to fail at once; the symlink there was nothing new (and I think the last line you had me remove was also representing that symlink)

wfeldt:
The line would create the same symlink again. It's not necessary. I was wondering why it worked before and I think it did because update-ca-certificates (which is run at the end) might have created the pem dir.

wfeldt:
Thanks!

PR #51, opened 2015-05-06
Ruby instalation:

DimStar77:
the actual binaries are called *.ruby2.1; reflect this in the created
symlinks

testrb no longer exists in ruby 2.2, remove it (the logic might need to
be enhanced to only create symlinks to existing targets).


DimStar77:
This should, to my understanding, help the failure currently observed in staging H
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:H/installation-images-openSUSE/standard/x86_64

PR #53, opened 2015-06-10
Switch off Linemode for PowerPC64 architectures

k0da:
Switch off Linemode for ppc64 both BE and LE variants.

Signed-off-by: Dinar Valeev <dvaleev@suse.com>

### DIFF
--- a/data/initrd/theme.file_list
+++ b/data/initrd/theme.file_list
@@ -45,7 +45,7 @@ if liveeval
   e echo "LiveConfig:	suselive.900" >>linuxrc.config
 endif
 
-if arch eq 'ppc' || arch eq 'ppc64' || arch eq 'ppc64le' || arch eq 's390' || arch eq 's390x'
+if arch eq 'ppc' || arch eq 's390' || arch eq 's390x'
   e echo "Linemode:	1" >>linuxrc.config
 endif
 


mvidner:
Thanks. It seems to make sense to me as a non-expert. But I'd like to take an opportunity to document things a bit:

From your usage it is evident that ppc, ppc64 and ppc64le are NOT subclasses of one another. In other places where they are not used together this may get overlooked. Let's add some explanation to the README.

k0da:
Well, ppc which is 32bit could be installed on older hardware. Where Linemode would be helpfull. I'm more interested in 64bit architectures and target Server hardware. Current hardware have no problems running in standard mode.

There was exception running 64bit on iSeries, which is no longer supported scenario. 

So I left 32bit for sake of compatibility.

k0da:
@mvidner I don't think it makes sense to document any particular platform, since the rest are not documented.

kobliha:
Side note: Documentation have to `start` somehow, `not being documented at all` doesn't sound like a reason for `not documenting ever`.

k0da:
@kobliha agree. But I don't think this should be addressed in this PR

k0da:
Currently it blocks Tumbleweed release:
https://openqa.opensuse.org/tests/67892

wfeldt:
@k0da: if you have a bug number for this, please mention it in the commit log. (IIRC there was at least one). It helps tremendously to get the changes trough the sp1 review process.

k0da:
@wfeldt I added bsc number into commit message

wfeldt:
Sorry for this but could you please move the bsc number into the first line? Only that line is used for the obs changelog entry.

k0da:
updated

PR #54, opened 2015-06-16
fix module dependency ordering (bsc #934580)

wfeldt:
Module ordering in modules.dep implies that you load the modules from right to left. In our config it's from left to right, so we have to revert the order at some point.

Strangely enough that rarely really matters so this remained unnoticed so far.


jreidinger:
LGTM

jreidinger:
I think it make sense to load from left to right as it is more nature in english

wfeldt:
It depends on how you interpret it. 'a b c' means a depends on b depends on c. So load c, then b, then a.

PR #55, opened 2015-06-26
Removed creating links to /usr/lib/YaST2/startup/YaST2.First-Stage

kobliha:
- FATE#317637
- /sbin/yast (/usr/bin/yast) calls this script itself
- also removed removing /usr/bin/yast
- This is for simpler dud=yast2.rpm - no special changes needed anymore (except yast.ssh)

See also https://github.com/yast/yast-yast2/pull/375

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -254,9 +254,10 @@ yast2:
   /usr
   /var
   r /usr/lib/YaST2/servers_non_y2/ag_cron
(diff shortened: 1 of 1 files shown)


kobliha:
@wfeldt No, not yet, I've just merged the PR so it will take some time. But it will be there (PR #56 is for SLE 12)

wfeldt:
Ok, here we go...

kobliha:
Thanks Steffen :)

PR #58, opened 2015-06-29
also create /dev/random and /dev/urandom in the initrd

msmeissn:
needed when we operate in FOPS mode.


kobliha:
Marcus, thanks for the patch. I hope it will be accepted to build even without so called "ID". Installation images use auto-submission to I/OBS.

PR #62, opened 2015-07-20
keep most of initrd in a squashfs image also on s390x (bsc #937012)

wfeldt:


jreidinger:
LGTM

wfeldt:
To explain it a bit:

the initrd contains cracklib-dict-small (via package deps). To be able to replace it via 'extend' the files
should be symlinked (extend does not replace real files). This can be achieved with the patch as then
the initrd files (most of them) are put into squashfs images (to save space). 

Another option would have been to remove cracklib-dict-small from initrd but I don't know what
side-effects this will have as the package is required (by pam).

ancorgs:
Shouldn't that nice explanation be somewhere else apart from this volatile pull request?

wfeldt:
well, I think github keeps them indefinitely...

PR #68, opened 2015-10-23
tell udevd not to set mount flags (bsc #937237)

wfeldt:
The rescue system does not have a mountpoint at its root node (it runs
chroot'ed). This will make udevd fail when it tries to apply a
'mount --make-shared' to it.


k0da:
Tested

aschnell:
LGTM

PR #73, opened 2015-11-20
Test 6

wfeldt:


mvidner:
What is the difference between "Boot from Hard Disk" and "Boot Linux System"? Even when I can read the code implementing it I don't understand what the latter does.

wfeldt:
... that's why it's in a submenu...

The 1st boots the hard disk, the 2nd boots an installed linux system.
Suggestions for a more descriptive label?

mvidner:
how does 'Boot Linux System' know which partition to boot? /boot/x86_64/loader/linux is on the dvd, right?

wfeldt:
It's extremely clever. :-)

wfeldt:
maybe we should record the reviews after all...

wfeldt:
to answer your question: linuxrc looks for an installed system

mvidner:
Aaah. Maybe "Boot a Linux System via Linuxrc" would be more informative then? Anyway, LGTM now.

PR #77, opened 2016-01-20
also include VERSION files for ruby gems

wfeldt:


jreidinger:
LGTM

kobliha:
@wfeldt I'd like to see some information in the changelog (generated from the commit) - it says `what` you have done, but it doesn't say `why`. We might be missing the information when changing it later. The current inst-sys is so fine-tuned that it's quite fragile to change it without knowing a lot of information, but these pieces of information are not anywhere.

kobliha:
Reported as https://bugzilla.suse.com/show_bug.cgi?id=962723

PR #80, opened 2016-01-26
rework product/version handling to use the values from /etc/os-release

wfeldt:
This removes the related static entries from etc/config and ensures
version and product info is always up-to-date.

Also, this implements the new driver update directory naming for
Tumbleweed ('x86_64-tw') and Leap ('x86_64-leap42.1').

### DIFF
--- a/etc/os_sample.txt
+++ b/etc/os_sample.txt
@@ -1,0 +1,69 @@
+== opensuse ==
+
+NAME=openSUSE
+VERSION="13.2 (Harlequin)"
+VERSION_ID="13.2"
+PRETTY_NAME="openSUSE 13.2 (Harlequin) (x86_64)"
+ID=opensuse
+ANSI_COLOR="0;32"
+CPE_NAME="cpe:/o:opensuse:opensuse:13.2"
+BUG_REPORT_URL="https://bugs.opensuse.org"
+HOME_URL="https://opensuse.org/"
+ID_LIKE="suse"
+
+== tumbleweed ==
+
+NAME=openSUSE
+VERSION="Tumbleweed"
+VERSION_ID="20160126"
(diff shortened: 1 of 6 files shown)


etc/os_sample.txt
```diff
@@ -0,0 +1,69 @@
...
+BUG_REPORT_URL="https://bugs.opensuse.org"
+HOME_URL="https://opensuse.org/"
+ID_LIKE="suse"
+
+== sles12 ==
+
+NAME="SLES"
+VERSION="12"
+VERSION_ID="12"
+PRETTY_NAME="SUSE Linux Enterprise Server 12"
```
imobach: NP: maybe we could use the 'new' PRETTY_NAME? (SUSE Linux Enterprise 12)

imobach: Never mind: I saw that you're distinguishing between server and desktop in this sample file.

wfeldt: The file is just a collection of current os-release files. It's purely informational.

lib/ReadConfig.pm
```diff
@@ -826,6 +828,57 @@ sub show_package_deps
...
+    }
+  }
+  else {
+    return;
+  }
+
+  # build nice product name
+
+  my $product = $config{PRETTY_NAME};
+  $product =~ s/\s*\([^)]*\)$//;
```
imobach: np: Some clarification about what's the intention would help when another one has to read the code.

imobach:
Just a minor np; otherwise, LGTM.

lib/ReadConfig.pm
wfeldt: Ok, in this case, strip the '(ARCH)' post fix, if there's any - see sample file.

PR #82, opened 2016-01-28
don't chainload shim for now (bsc #963919)

wfeldt:
There seems to be a problem with shim chainloading (another) shim
atm; so avoid it.


wfeldt:
Note that this is a workaround and it should be removed if bsc #963919 has been fixed.

kobliha:
LGTMed `out of GitHub` by this comment: https://bugzilla.suse.com/show_bug.cgi?id=963919#c3

PR #83, opened 2016-02-03
adjust product/version handling

wfeldt:
The existing implementation would produce unexpected driver update layouts
for SLES derived products. So we adjust the approach a bit and make sure
only the documented driver update dist tags survive (sles, sled, leap, tw).
 On branch test_20


imobach:
LGTM

PR #85, opened 2016-02-11
giving up on rubygem packaging (bsc #966145)

wfeldt:
rubygem packaging is just so crappy I'm giving up now and include the whole gems, even if that means wasting disk space.


imobach:
LGTM

ancorgs:
Too late but I was going to suggest to leave a comment in place of the removed lines, so we are not tempted to try to tame rubygems again in the future.

PR #86, opened 2016-02-17
s390x: add support for HMC (fate #318899)

wfeldt:


mvidner:
Yes, this is short but it is absolutely incomprehensible, unless you know where to look. Fortunately it can be easily fixed by adding a couple of URL spiced comments...

data/initrd/scripts/url/hmc/mount
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
```
mvidner: \# Implement the 'hmc' URL scheme for linuxrc
\# See https://github.com/openSUSE/linuxrc/blob/a72ccf44eb19b6d55bc54058876fa5f64021eee6/user_defined_url_schemes.txt

data/initrd/scripts/url/hmc/menu
```diff
@@ -0,0 +1 @@
+HMC
```
mvidner: HMC may be the usual term, but can we hint at the meaning for the uninitiated?
> HMC (Hardware Magnification Controller)

data/initrd/scripts/url/hmc/mount
```diff
@@ -0,0 +1,16 @@
...
+
+set -e
+
+echo "url_dir = '$url_dir'"
+
+modprobe hmcdrv
+
+if [ -d $url_dir ] ; then
+  hmcdrvfs $url_dir
```
mvidner: Is the program already included? s390utils-hmcdrvfs.rpm does not seem to be mentioned anywhere

data/initrd/scripts/url/hmc/menu
wfeldt: Too long.

HMC is perfectly clear to any s390 enthusiast.

kobliha: In fact C == Console, but I'd rather see it too long as not everyone using it IS actually an enthusiast (QA, ourselves, ...)

wfeldt: You have to prepare your machine specifically to make use of the HMC option. So you'd hardly miss the entry.

markkp: What's the reason for saying it's "too long?"  Would
HMC (Hardware Mgmt. Cons)
or
HMC (Hdw. Mgmt. Console)
be better? I tend to agree that it would be better to give _some_ clue here.

data/initrd/initrd.file_list
```diff
@@ -569,6 +570,10 @@ x etc/syslog.conf etc
 
 # additional scripts for linuxrc
 x scripts /
+# hmc exists only on s390
```
markkp: Not entirely true (Power systems have them), but might be true enough for our purposes. Power system users don't have the same problem getting access to a DVD.

data/initrd/initrd.file_list
```diff
@@ -569,6 +570,10 @@ x etc/syslog.conf etc
...
+if !exists(s390-tools-hmcdrvfs)
```
markkp: I'm not sure if I like this compared to the previous check for an arch of s390 or s390x. What's the benefit of doing it this way?

data/initrd/scripts/url/hmc/mount
```diff
@@ -0,0 +1,40 @@
...
+  exit 1
+fi
+
+# the rewritten repo location will not be available in the installed system,
+# so we inject a script to remove our repo from zypper
+#
+# linuxrc uses 3-digit numbers to store diver updates; so actually any
+# number != 3 digits will do to avoid conflicts
+
+mkdir -p /update/4387/install
```
markkp: Where did the number 4387 come from? (Mostly just curious)

data/initrd/scripts/url/hmc/mount
```diff
@@ -0,0 +1,40 @@
...
+
+set -e
+
+echo "url_dir = '$url_dir'"
+
+modprobe hmcdrv
+
+if [ -d "$url_dir" ] ; then
+  hmcdrvfs "$url_dir"
+
```
markkp: I think we need to insert some checking in here to see if
1. The mount was successful
2. There is actually _something_ under the mountpoint
2a. It looks like an installation tree

Using lshmc, we can actually do 2 and 2a after the modprobe and _before_ the hmcdrvfs command

wfeldt: Note the 'set -e'

It's not the task of this script to verify anything. linuxrc will do that. This script is only there to provide the mount.

data/initrd/scripts/url/hmc/mount
wfeldt: > any number [...] will do

It's some random number.

kobliha: So not to use real tools for making it really random, e.g., `mktemp`?

data/initrd/scripts/url/hmc/menu
wfeldt: 'Hardware Management Console' is even longer than 'Digital Video Disk', yet we use 'DVD'.

My impression was that HMC is a well-known term for s390 admins. But you're from the s390 side. If you feel more comfortable with 'Hardware Management Console', then so be it. 'Hdw. Mgmt. Console', however, looks really awkward to me.

> Also, is the lack of a line-end character here of any importance?

The file holds the string for the menu entry, no newline needed. So, why waste a byte just because github shows a red symbol?

data/initrd/initrd.file_list
wfeldt: It ultimately depends on the tools, not the arch. We can have the hmc url scheme once the hmcdrvfs command exists.

data/initrd/initrd.file_list
wfeldt: Ok, so if there's any chance to see this menu entry ever also on ppc, this might be an argument for the long menu name.

wfeldt:
To get it going, I'm submitting.

data/initrd/scripts/url/hmc/menu
markkp: >>> On 2/18/2016 at 04:25 AM, Steffen Winterfeldt <notifications@github.com> wrote:

> My impression was that HMC is a well-known term for s390 admins. But you're 
> from the s390 side. If you feel more comfortable with 'Hardware Management 
> Console', then so be it.

It is a well-known term, but I tend towards being more explanatory than not, when possible.  My comfort level was more in consideration of others that might not be as familiar.

PR #89, opened 2016-02-22
Update initrd.file_list according to the sysvinit change

nilxam:
Now showconsole is splits from sysvinit as the separate package called blog.

Per https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:C/installation-images-openSUSE/standard/x86_64 , I hope this is the right fix, please review.

Ref. https://build.opensuse.org/request/show/356283 and https://build.opensuse.org/request/show/360535

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -260,9 +260,11 @@ ntfs-3g:
 ?kexec-tools: nodeps
   /usr/sbin/kexec
 
-sysvinit-tools: nodeps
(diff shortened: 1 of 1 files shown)


nilxam:
ok, let's closing this one.

wfeldt:
see https://github.com/openSUSE/installation-images/pull/90

wfeldt:
I wonder what there isn't a libblogger package, according to package policy?

PR #94, opened 2016-03-07
More detailed listing of ifaces provided in ssh installation.

mchf:
bnc#956473

### DIFF
--- a/data/root/etc/inst_setup
+++ b/data/root/etc/inst_setup
@@ -219,6 +219,14 @@ EOF
   cp /etc/issue /etc/motd
   echo -e "Run yast.ssh to start the installation.\n" >>/etc/motd
 
+  # print more detailed list of ifaces using a function from yast2-installation
+  if [ -e /usr/lib/YaST2/startup/common/network.sh ] ; then
+    echo "Active interfaces:"
+
+    . /usr/lib/YaST2/startup/common/network.sh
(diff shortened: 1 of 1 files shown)


mchf:
I was thinking of it whole weekend. In my POV 90% of users are satisfied with the message as is. 

However, there can still be some which would like to see alternatives. The function which is used tries to limit listing (doesn't show unconfigured ifaces, loopback), but to be honest I really don't know if that is useful. May be a question for PjM or UX?

wfeldt:
We actually already show an ip list, gathered from wicked. It just happens to be invisible, as plymouth
erases the screen afterwards (but you see it with plymouth=0). :-(

Maybe we should just move the 'plymouth_off' line a bit up in the script (to before the /tmp/hostips line)?

mchf:
thanks for tip, will try

kobliha:
PLS see the respective Trello card, the list of ifaces - as it is shown now - is only confusing, containing a lot of unneeded information. Let's show only those items, that are necessary. And aligned and sorted.

mchf:
@wfeldt Do you have some more comments regarding this PR? Pls see corresponding trello card for additional discussion. Thanks

mchf:
Device listing was updated according UX comments here: https://github.com/yast/yast-installation/pull/364

PR #96, opened 2016-03-22
Build inst-sys extension for libstoragemgmt (fate#320272)

jsrain:


data/root/config
```diff
@@ -1,4 +1,4 @@
-root:			common root cracklib-dict-full.rpm bind ?yast2-trans-<lang>.rpm
+root:			common root libstoragemgmt cracklib-dict-full.rpm bind ?yast2-trans-<lang>.rpm
```
wfeldt: This means libstoragemgmt will always be loaded, which is not your intention, iirc. Just ,leave the line unchanged.

jsrain: OK, but then I'm perhaps missing something about cracklib-dict-full.rpm - it is loaded later by YaST...

wfeldt: No, you're not missing anything. It's always loaded...

PR #99, opened 2016-04-28
allow user to switch efi grub2 console to text mode

wfeldt:


wfeldt:
agraf has tested this

PR #100, opened 2016-05-02
adjust config to new 'hiddenentry' feature (bsc #976836)

wfeldt:
It was changed from 'menuentry --hidden' to 'hiddenentry'.


wfeldt:
checked by agraf

PR #101, opened 2016-05-04
sanitize hostname handling (fate #319639)

wfeldt:
Note that there are also linuxrc changes that go along with this.

### DIFF
--- a/data/root/etc/inst_setup
+++ b/data/root/etc/inst_setup
@@ -54,16 +54,9 @@ export YAST2_SSH=false
 unset SSH_FAILED
 stty sane 2>/dev/null
 
-# set hostname and domainname
-hostip_from_wicked >/tmp/hostips
-
-hostname=$(awk ' /^Hostname:/ { print $2 }' < /etc/install.inf)
-[ -n "$hostname" ] && hostname $hostname
-hostname=`hostname`
-
-domain=$(awk ' /^Domain:/ { print $2 }' < /etc/install.inf)
-[ -z "$domain" ] && domain=local
-domainname $domain
+# get hostname & hostips
+hostip_from_wicked >/tmp/host_ips 2>/tmp/host_name
+host_name=`cat /tmp/host_name`
 
 #
 # a few files should be restored when installation has completed if we
@@ -186,7 +179,7 @@ function start_shell() {
   bash -l
 }
 
-[ -f /tmp/hostips ] && cat /tmp/hostips
+[ -f /tmp/host_ips ] && cat /tmp/host_ips
 
 [ "$START_SHELL" ] && start_shell
 
@@ -197,7 +190,7 @@ if grep -qi "^VNC:.*1" /etc/install.inf ; then
   if test "$ec" = "0" ; then
     (
     sleep 3
-    /usr/bin/slptool register "service:YaST.installation.suse:vnc://${hostname}:5901"
+    /usr/bin/slptool register "service:YaST.installation.suse:vnc://${host_name}:5901"
     ) &> /tmp/slptool_register.txt &
   else
     echo "slpd returned with exit code $ec, VNC will not be announced"
@@ -212,7 +205,7 @@ ec=
 if [ "$YAST2_SSH" = "true" ] ; then
   cat <<EOF
 
-***  login using 'ssh -X root@${hostname}'  ***
+***  login using 'ssh -X root@${host_name}'  ***
 ***  run '${yast}.ssh' to start the installation  ***
 
 EOF
--- a/data/root/etc/inst_setup
+++ b/data/root/etc/inst_setup
@@ -55,7 +55,7 @@ unset SSH_FAILED
 stty sane 2>/dev/null
 
 # get hostname & hostips
-hostip_from_wicked >/tmp/host_ips 2>/tmp/host_name
+hostip_from_wicked /tmp/host_ips /tmp/host_name
 host_name=`cat /tmp/host_name`
 
 #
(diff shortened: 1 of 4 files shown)


wfeldt:
https://github.com/openSUSE/linuxrc/pull/108

data/root/etc/inst_setup
```diff
@@ -197,7 +190,7 @@ if grep -qi "^VNC:.*1" /etc/install.inf ; then
   if test "$ec" = "0" ; then
     (
     sleep 3
-    /usr/bin/slptool register "service:YaST.installation.suse:vnc://${hostname}:5901"
+    /usr/bin/slptool register "service:YaST.installation.suse:vnc://${host_name}:5901"
```
mchf: ```hostip_from_wicked``` prints hostname including ```\n``` at the end. Isn't it a problem (e.g.) here?

data/root/hostip_from_wicked
```diff
@@ -1,5 +1,13 @@
 #! /usr/bin/perl
 
+# helper script that
+#
+# (1) writes a list of assigned IP adresses to STDOUT, and
+#
+# (2) writes the FQDN or, if not available, the first IP to STDERR
```
mchf: returning value through STDERR seems a bit obscure to me.

data/root/etc/inst_setup
wfeldt: no, shell swallows white space

data/root/hostip_from_wicked
wfeldt: true, but I documented it...

If you think this a serious problem, I could go and rewrite the script.

mchf: as this is about implementing new feature and not about a hotfix I'd go for more clean implementation.

data/root/etc/inst_setup
mchf: ehm, ok. So why do you explicitly append it? I'm just curios, not blocker for me ;-)

wfeldt: No idea; the code is not by me. If you say it's not necessary, I'm fine with another line.

data/root/etc/inst_setup
```diff
@@ -55,7 +55,7 @@ unset SSH_FAILED
 stty sane 2>/dev/null
 
 # get hostname & hostips
-hostip_from_wicked >/tmp/host_ips 2>/tmp/host_name
+hostip_from_wicked /tmp/host_ips /tmp/host_name
```
mchf: ehm, is it possible to split / refactor this into two functions / helper scripts to have just ```hostip_from_wicked``` and e.g. ```hostname_from_wicked```?

Moreover, I'd fix the behavior when ```host_name``` sometimes contains hostname and sometimes first ip.

wfeldt: Why can't I extract two things from a single wicked run?

host_name is either a fqdn or the ip if no dns is available. I think it's named 'fqhn' sometimes. What's so bad about it?

mchf: > Why can't I extract two things from a single wicked run?

Common practice is to have one method to do just one thing. In the current state I'd expect that the function / script returns just one IP ... In reality it writes a list of ips into a file (with kind of header prepended) and hostname into another file.

> host_name is either a fqdn or the ip if no dns is available. I think it's named 'fqhn' sometimes. What's so bad about it?

well, so far we has been talking about fqdn. So at least documentation should be fixed.

mchf: @mvidner As I'm not strong in perl could you pls also review? Thanks

wfeldt: > Common practice is to have one method to do just one thing

I'd follow you for ruby methods or oo in general, but not shell scripts. If the name irks you, we can rename
the script to something else.

> So at least documentation should be fixed.

It does exactly what is documented.

data/root/hostip_from_wicked
```diff
@@ -1,5 +1,17 @@
...
+# (1) writes a list of assigned IP adresses to ARG1, and
+#
+# (2) writes the FQDN or, if not available, the first IP to ARG2
+#
+# The address in (2) should be something we can be reached by via network.
+#
+# usage:
+#
+# hostip_from_wicked ARG1 ARG2
+
```
mvidner: Thanks for the docs!

I would also add a sample wicked output, so that the parser loop makes more sense to the future maintainer:
```console
# wicked show all
lo              up
      link:     #1, state up
      type:     loopback
      config:   compat:suse:/etc/sysconfig/network/ifcfg-lo
      leases:   ipv4 static granted
      leases:   ipv6 static granted
      addr:     ipv4 127.0.0.1/8 [static]
      addr:     ipv6 ::1/128 [static]

ens7            up
      link:     #2, state up, mtu 1500
      type:     ethernet, hwaddr 52:54:00:05:0a:2c
      config:   compat:suse:/etc/sysconfig/network/ifcfg-ens7

eth1            up
      link:     #3, state up, mtu 1500
      type:     ethernet, hwaddr 52:54:00:d7:57:86
      config:   compat:suse:/etc/sysconfig/network/ifcfg-eth1
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp requesting
      addr:     ipv4 10.230.84.112/18 [dhcp]
      route:    ipv4 default via 10.230.63.254
```

data/root/etc/inst_setup
mchf: >> Common practice is to have one method to do just one thing

> I'd follow you for ruby methods or oo in general, but not shell scripts. If the name irks you, we > can rename
> the script to something else.

then rename. Thanks

data/root/hostip_from_wicked
wfeldt: fair enough; added

mvidner:
LGTM

PR #105, opened 2016-05-19
Ruby debugger support (FATE#318421)

lslezak:
- Fix `byebug` symlinks (`update-alternatives` cannot be used)
- Additionally remove the not needed `cache/*.gem` files to save some space

After the space optimization the `root` image is smaller by ~60kB, so even after adding the Ruby debugger we still have a smaller image than before.


wfeldt:
ok

PR #106, opened 2016-05-24
module.config: add macsec

jirislaby:
This is a module added in 4.6 and similar to macvlan, macvtap.


jirislaby:
Fixes bnc#981113

PR #107, opened 2016-05-30
adjust to update-alternative package change

wfeldt:
There's a symlink alternatives -> update-alternatives now in the package which must also be cleaned up.


lslezak:
LGTM

PR #108, opened 2016-06-10
Add xen-tools-domU on x86 to get its udev rules (bsc#979002)

olafhering:
For  TW and SLE12.
Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -95,6 +95,7 @@ tar:
 util-linux-systemd:
 vlan:
 which:
+?xen-tools-domU:
 
 if theme eq 'SLES'
   sles-release:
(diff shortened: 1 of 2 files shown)


wfeldt:
If you are after udev rules wouldn't it be better added to the initrd?

olafhering:
I have updated the reqest

wfeldt:
Thanks; but you'll still want it in the rescue system, too.

olafhering:
Why is it required for rescue? Does the initrd or rescue load the drivers and as such process the events?

wfeldt:
initrd loads the drivers initially, but rescue is a complete system, running with init(1); so everything udevd has done, except for the module loading, is redone. In contrast, root is just an extension run in the same environment as the initrd.

So, if your udev rules just trigger some modules, you are ok putting them into the initrd alone. But if you also create some symlinks in /dev, you'll also need to adjust the rescue system.

olafhering:
Ok, now I restored the rescue part.

PR #109, opened 2016-06-15
change how rubygem-byebug is added to inst-sys

wfeldt:
It can happen that the project we build against has several versions of ruby
gems. To avoid that the gem is installed more than once, add it as TEMPLATE
to file_list, so that it's dragged in only via package dependencies.


lslezak:
LGTM, thanks!

PR #111, opened 2016-06-17
handle build id in Makefile

wfeldt:
This used to be done in the spec file but this has the disadvantage that it
doesn't work in local builds (via a simple 'make').

(Since yast now relies on the existence of .instsys.config, it's no longer
optional.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ GITDEPS := $(shell [ -d .git ] && echo .git/HEAD .git/refs/heads .git/refs/tags)
 VERSION := $(shell $(GIT2LOG) --version VERSION ; cat VERSION)
 BRANCH  := $(shell [ -d .git ] && git branch | perl -ne 'print $$_ if s/^\*\s*//')
 PREFIX  := installation-images-$(VERSION)
+BUILD_ID := $(shell [ -f .build_id ] || bin/build_id > .build_id ; cat .build_id)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ GITDEPS := $(shell [ -d .git ] && echo .git/HEAD .git/refs/heads .git/refs/tags)
 VERSION := $(shell $(GIT2LOG) --version VERSION ; cat VERSION)
 BRANCH  := $(shell [ -d .git ] && git branch | perl -ne 'print $$_ if s/^\*\s*//')
 PREFIX  := installation-images-$(VERSION)
+BUILD_ID := $(shell [ -f .build_id ] || bin/build_id > .build_id ; cat .build_id)
```
imobach: NP: Maybe you could use `tee` (`bin/build_id | tee .build_id`)?

imobach:
LGTM. Just a minor NP (feel free to ignore it) :)

mvidner:
Is there a bug id? User-visible symptoms of the bug that this is fixing?

Makefile
wfeldt: No, you're missing the point of the '||'. :-)

wfeldt:
No, this hits only the developer.

Makefile
imobach: Yes, you're right :)

PR #116, opened 2016-07-11
Drop some unused packages from the root image

lslezak:
ignore dependencies:
- yast2-devtools
- yast2-buildtools

removed packages:
- perl-Bootloader
- perl-XML-Parser
- perl-XML-Simple

## Result

As the result these packages were removed from the inst-sys (compared the `.packages.root` files in inst-sys):

- perl-Bootloader
- perl-XML-Parser
- perl-XML-SAX
- perl-XML-SAX-Base
- perl-XML-SAX-Expat
- perl-XML-Simple
- perl-XML-Writer

The size of the `root` image decreased by ~320kB.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -98,6 +98,14 @@ yast2-schema: ignore
 # already in initrd
 ca-certificates-mozilla: ignore
 filesystem: ignore
+# not needed during installation, added by the
+# "perl-Bootloader < kexec-tools < kdump < yast2-kdump" dependency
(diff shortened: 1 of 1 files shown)


wfeldt:
This will likely not do what you intend. 'ignore' prevents the whole package from being installed. You might be looking for 'nodeps'.

lslezak:
Thanks Steffen, I have updated the file accordingly.

wfeldt:
lgtm

PR #117, opened 2016-07-12
reduce instsys size by removing files already in initrd

wfeldt:
Here's how it works:

`common_tree --dst xxx foo bar`

will create 3 directories `xxx/1`, `xxx/2`, and `xxx/c` which contain `all in foo but not in bar`, `all in bar but not in foo`, and `everyting both in foo and bar`, respectively.

So, in this patch, `tmp/root+initrd/2` contains everything in the root image that's not already in the initrd.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ boot: base
 	theme=$(THEMES) image=boot fs=dir bin/mk_image
 
 root: base
-	theme=$(THEMES) libdeps=root image=root bin/mk_image
+	theme=$(THEMES) libdeps=root,initrd image=root bin/mk_image
 
 rescue: base
 	theme=$(THEMES) libdeps=rescue image=rescue bin/mk_image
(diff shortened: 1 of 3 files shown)


lslezak:
Um... the patch is bit tricky for me. I can see some added commands, but for me it's not obvious that the result of these commands is removing the duplicates.

Maybe it's just because I'm not familiar with the code, but I'd appreciate adding some comments which describe where is the trick and what's the result.

lslezak:
Thanks for the details!

Could you put that info into the code or to some documentation? it should be at some more "permanent" place than in a PR description...

jreidinger:
I think ideal place is the comment at top of https://github.com/openSUSE/installation-images/blob/master/bin/common_tree

jreidinger:
Thanks for placing comment there. Now I think it is much better

jreidinger:
LGTM

lslezak:
LGTM as well, thanks!

PR #118, opened 2016-07-14
Adjust the OOM killers score for haveged (bsc#974601)

lslezak:
The OOM killer should prefer killing the haveged daemon first, the installation can be finished without it.
Haveged only generates the entropy for kernel, kernel does that itself but slowly.

According to the log in the bug haveged takes about 4MB memory - not much but maybe be that after killing it the installation can be finished successfully.


wfeldt:
lgtm

PR #120, opened 2016-07-20
Include perl-XML-Simple again in inst-sys (bsc#989702)

lslezak:
`perl-XML-Simple` requires several other Perl libraries and was a good candidate for the space reduction. Originally I grepped for `.anyxml` usage in YaST and thought it was not needed during installation.

Unfortunately it turned out that the `OneClickInstallStandard.GetRepositoriesFromXML()` call, which is part of the one click installer, uses `.anyxml` and is used in installation to parse the extra repository list in openSUSE.


jreidinger:
LGTM

PR #121, opened 2016-07-20
exclude kmod-compat (bsc #989706)

wfeldt:
We ignore kmod as it's in the initrd already. Unfortunately kmod-compat
creeps in on ppc64le as dependency of powerpc-utils. So, exclude it
explicity as well.


jreidinger:
LGTM

PR #123, opened 2016-07-28
adjust to recent package changes

wfeldt:
python-setuptools is really not needed, systemd no longer has libs, and
pam-modules no longer has anything in /sbin


jreidinger:
LGTM

PR #124, opened 2016-07-29
update module.config to match 4.7

jirislaby:
mv88e6xxx_drv was renamed to mv88e6xxx.

gtp tunelling was added -> put is along pppoe and similar.

Signed-off-by: Jiri Slaby <jirislaby@gmail.com>


wfeldt:
Jiri, please only add, don't remove entries. This way we also build with older kernels. Entries for non-existent modules will be auto-removed during build.

jirislaby:
Done :).

wfeldt:
thanks

PR #125, opened 2016-08-04
Make pam-modules optional

lnussel:
In Factory pam-modules only contains legacy modules that are not
needed by default anymore. Therefore pam-modules won't be available
in buildroot by default anymore.


lnussel:
needed for https://build.opensuse.org/package/live_build_log/openSUSE:Leap:42.2:Staging:A/installation-images-openSUSE/standard/x86_64

lnussel:
so?

PR #126, opened 2016-08-09
add zypp config to initrd to ensure it's writable (bsc #967828)

wfeldt:
/etc in initrd is guaranteed to be writable. So put zypp config here.


lslezak:
LGTM, thanks!

PR #134, opened 2016-10-14
Added susehmc.ins and parmfile.hmc files to allow installation from t…

markkp:
…he DVD reader in the HMC without a network connection. (bsc#1002974)


wfeldt:
Mark, please also adjust install.s390x and submit against sle12-sp2 branch.

PR #138, opened 2016-10-18
fix susehmc.ins installation (bsc #1002974)

wfeldt:


wfeldt:
>>>> On 10/17/2016 at 10:41 AM, Steffen Winterfeldt <snwint@suse.de> wrote:
>> BTW, there is a suse.ins in / and in /boot/s390x/ on the CD. Which one is
>> relevant?

> From what I can tell, they're both relevant.  The one in / has the same contents, but has
> relative paths into boot/s390x.

Well, you wrote it's needed in both locations.

In any case, the path patching looks necessary to me. `/susehmc.ins` would need the full path inside, or not?

markkp:
I didn't say it was _needed_ just that it could be considered relevant.  From my perspective, we could get rid of boot/s390x/suse.ins, but I don't know if anyone else "out there" is depending on it being there.  So while I don't care to remove /boot/s390x/suse.ins, I'm not particularly interested in adding /boot/s390x/susehmc.ins to further confuse people in the future.

Dear $DIETY, I'm becoming incompetent.  Yes /susehmc.ins needs to have boot/s390x/ in column 1 of the file names.  I was going to try and have someone burn a real DVD and put it into the zEC12 for testing, but we haven't gotten to the point where they can do that yet.  Then you, at least, figured out that it wouldn't have worked anyway.  Sigh.  I'm beginning to wonder if I can be trusted at all. I've submitted another pull request with the correct(?) contents.

wfeldt:
obsoleted by https://github.com/openSUSE/installation-images/pull/139

PR #140, opened 2016-10-21
update module.config to match 4.9

jirislaby:
4.9 added ledtrig-usbport, so add it to the list along with other led
drivers.

Signed-off-by: Jiri Slaby jslaby@suse.cz


jirislaby:
ping

PR #142, opened 2016-11-09
stick to grub default colors (bsc #1005899)

wfeldt:
The module config change in https://github.com/openSUSE/installation-images/pull/140 was not complete - it needs an additional entry.


imobachgs:
LGTM

PR #144, opened 2016-11-13
Doc fixes

dwaas:
See the rendered version here:

http://installation-images-dwaas.readthedocs.io/en/latest/


wfeldt:
Isn"t index.md a symlink to README.md atm? Can't we just git rm/mv things?

PR #145, opened 2016-11-18
removed local alias

dwaas:
the alias is used to get su privileges, but it's not required because we
run under root anyway.


wfeldt:
lgtm

PR #146, opened 2016-11-22
linuxrc's loghost scripts added FATE#318971

dwaas:
now the preferred logger is rsyslog so syslog was removed completely

It works, but it still needs refactoring:
1) removing ugly output during nscd start 

And more features:
1) reflect changes on rescue image as well
2) log more files

Implements https://fate.suse.com/318971

### DIFF
--- a/data/initrd/scripts/remote_log_setup
+++ b/data/initrd/scripts/remote_log_setup
@@ -1,0 +1,12 @@
+#! /bin/bash
+#USAGE: requires the LOGHOST variable to be set
+
+#set loghost
+echo "*.*@${LOGHOST}:514" > /etc/rsyslog.d/loghost.conf
+
+echo -n "starting rsyslogd (logging to /var/log/messages)..."
+if /usr/sbin/rsyslogd ; then
+  echo " ok"
+else
+  echo " failed"
+fi
--- a/data/initrd/etc/rsyslog.conf
+++ b/data/initrd/etc/rsyslog.conf
@@ -1,0 +1,92 @@
+# SUSE SPECIFIC rsyslog.conf used for Loghost capability in linuxrc
+###################################################################
+
+# provides --MARK-- message capability (every 1 hour)
+$ModLoad immark.so
+$MarkMessagePeriod      3600
+
+# provides support for local system logging (e.g. via logger command)
+$ModLoad imuxsock.so
+
+# reduce dupplicate log messages (last message repeated n times)
+$RepeatedMsgReduction   on
+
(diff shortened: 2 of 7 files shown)


data/initrd/scripts/remote_log_setup
```diff
@@ -0,0 +1,66 @@
+#! /bin/sh
+#USAGE: ./script_name $remote_log_host_ip
+
+#preconditions
+echo -n "linking mkdir"
+if  ln -s /usr/bin/mkdir /bin/mkdir ; then
+  echo " ok"
+else
+  echo " failed"
```
lslezak: Does it make sense to continue when something fails? I guess it will fail later anyway...

data/initrd/scripts/remote_log_setup
```diff
@@ -0,0 +1,66 @@
+#! /bin/sh
+#USAGE: ./script_name $remote_log_host_ip
+
+#preconditions
+echo -n "linking mkdir"
+if  ln -s /usr/bin/mkdir /bin/mkdir ; then
```
lslezak: This `if ... echo "ok" else echo "failed"` pattern is repeated several times, I think it could be shared in a function.

mvidner:
Thanks! Please add a reference to the bug or feature this is solving to the commit message, and to the top of the `remote_log_setup` script.

data/initrd/scripts/remote_log_setup
```diff
@@ -0,0 +1,66 @@
+#! /bin/sh
+#USAGE: ./script_name $remote_log_host_ip
+
+#preconditions
```
wfeldt: please do not patch our own system; add the link directly to the initrd via data/initrd/initrd.file_list

dwaas:
The latest version of the commit correctly forwards kernel messages to /dev/tty4.

NOTE: No checks are done to make sure that /dev/tty4 is present before the forwarding is setup.

data/initrd/scripts/remote_log_setup
```diff
@@ -0,0 +1,44 @@
+#! /bin/bash
+#Requires the LOGHOST variable to be set
```
wfeldt: Again, that's all not necessary. You don't really modify rsyslog.conf - it's always the same file! So just create the config you like and include the file directly in initrd.file_list replacing the default config.

dwaas: Given that the only "dynamic" piece of information is the loghost IP address, this will be stored in an additional file. Also we will ship a suse specific rsyslog.conf.

PR #147, opened 2016-11-28
adjust dud directory calculation to new tumbleweed os-release format (bsc #1012241)

wfeldt:
For whatever reason the /etc/os-release content has changed for tumbleweed
(compare tw entries in os_sample.txt). Adjust parser to new entries.


ancorgs:
lgtm

PR #155, opened 2017-01-05
update module.config to match 4.10

jirislaby:
4.10 added `scr24x_cs`, so add it to the list along with other unneeded
PCMCIA drivers.

kvm cannot be found by the scripts:
```
  error: no useful config for "kvm" (needed by kvm)
```
so hint it with a path.

### DIFF
--- a/etc/module.config
+++ b/etc/module.config
@@ -650,7 +650,7 @@ ioc3
 ipwireless
 iscsi_trgt
 joydev
-kvm
+kernel/arch/.*/kvm/kvm
 monreader
 msdos
 parport_ax88796
@@ -666,6 +666,7 @@ ppp_synctty
 pppoe
 prism2_usb
 pt
+scr24x_cs
 serial_cs
 slip
 tape_3590


wfeldt:
I'm wondering how kvm.ko comes into play. It's not mentioned in etc/module.list. So I guess some
other module requires it. I think we'd rather block that module instead.

jirislaby:
Only `kernel/drivers/gpu/drm/i915/gvt/kvmgt.ko` (a new driver in 4.10) depends on kvm, but that is not listed anywhere too.

Could you advise me, what to do?
```
# grep :.*kvm modules.dep
kernel/arch/x86/kvm/kvm-intel.ko: kernel/arch/x86/kvm/kvm.ko kernel/virt/lib/irqbypass.ko
kernel/arch/x86/kvm/kvm-amd.ko: kernel/arch/x86/kvm/kvm.ko kernel/virt/lib/irqbypass.ko
kernel/drivers/gpu/drm/i915/gvt/kvmgt.ko: kernel/drivers/vfio/mdev/mdev.ko kernel/drivers/vfio/vfio.ko kernel/drivers/gpu/drm/drm.ko kernel/arch/x86/kvm/kvm.ko kernel/virt/lib/irqbypass.ko
```

wfeldt:
I've had a look and it's dragged in by kernel/driver/gpu/.*. So maybe it is useful after all. ;-)

The solution here is just to remove the 'kvm' line.

wfeldt:
fixed as part of https://github.com/openSUSE/installation-images/pull/160

PR #157, opened 2017-01-12
zypp.conf is really in initrd, so modify it there

wfeldt:
Well, actually libzypp is **both** in initrd and root image but the first takes precedence.

So we also remove zypp config from root image to save the space.


jreidinger:
LGTM

PR #158, opened 2017-01-12
additional log files for FATE#318971

dwaas:
Loghost now sends also linuxrc, zypp's history, pk_backend_zypp,
zypper.log, pbl.log, /.packages.root.

All the files have distinct tags to allow filtering on the remote log
server

Asserted only that linuxrc.log  are sent. The rest should be consistent enough to hold.

Log files selected to be consistent with save_y2logs script.

### DIFF
--- a/data/initrd/etc/rsyslog.d/logged_files.conf
+++ b/data/initrd/etc/rsyslog.d/logged_files.conf
@@ -7,3 +7,39 @@ input(type="imfile"
       Tag="yast2_logs"
       Severity="notice"
       Facility="local0")
+
+#linuxrc
+input(type="imfile"
+      File="/var/log/linuxrc.log"
+      Tag="linuxrc"
+      Severity="notice"
+      Facility="local0")
+
+
+#zypp history
+input(type="imfile"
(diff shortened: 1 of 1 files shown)


data/initrd/etc/rsyslog.d/logged_files.conf
```diff
@@ -7,3 +7,46 @@ input(type="imfile"
...
+#pbl
+input(type="imfile"
+      File="/var/log/pbl.log"
+      Tag="pbl"
+      Severity="notice"
+      Facility="local0")
+
+#.packages.root
+input(type="imfile"
+      File="/.packages.root"
```
wfeldt: The file exists only after the inst-sys has been added but rsyslogd might be started earlier by linuxrc.
Does this work (rsyslogd sends the file once the symlink get created)?

dwaas: It does not work. It turns also out that since this is a static file present in the ISO, that's how the user is supposed to retrieve it (eventually).

The commit was amended removing the entry.

wfeldt:
lgtm

PR #165, opened 2017-01-27
prepare tree for tftp installations (fate #322275)

wfeldt:


shundhammer:
LGTM (for all as I can make sense of it)

PR #174, opened 2017-03-31
vim uses update-alternatives

wfeldt:
We don't want update-alternatives, so we'll have to create the symlink manually.


nilxam:
:+1:

PR #182, opened 2017-05-06
Add code for openssl version namespacing

scarabeusiv:
This is currently being tested in staging:J.
Used for SLE15, won't be backported.

PR #183, opened 2017-05-10
anticipate openssl repackaging

wfeldt:
openssl has been split into openssl (dummy) and openssl-VERSION.

replaces https://github.com/openSUSE/installation-images/pull/182

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -359,7 +359,11 @@ cracklib: nodeps
   x etc/ssh /lib
   x ../rescue/mount-rootfs-and-do-chroot.sh /bin
 
-openssl:
+if exists(openssl, /usr/share/ssl)
+  openssl:
+else
+  openssl-*:
+endif
   /usr/share/ssl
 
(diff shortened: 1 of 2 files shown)


data/root/root.file_list
```diff
@@ -359,7 +359,11 @@ cracklib: nodeps
   x etc/ssh /lib
   x ../rescue/mount-rootfs-and-do-chroot.sh /bin
 
-openssl:
+if exists(openssl, /usr/share/ssl)
+  openssl:
```
lslezak: (note: https://github.com/openSUSE/installation-images/blob/master/doc/files.md#conditional-sections explains the condition syntax)

Um, could we use a simple `openssl*` here? That would match both `openssl-*` and `openssl`. Or not?

wfeldt: The directory is not on both and you get a warning logged. I wanted to avoid that.

lslezak: I see, good point, thanks!

PR #184, opened 2017-05-12
build 'real' SLED theme (bsc #1038838)

wfeldt:
Now that installation-images has become a multibuild package, we can build a
'real' SLED variant (one that's not just a copy of SLES).


ancorgs:
LGTM

PR #185, opened 2017-05-29
allow driver updates also to be applied to the rescue system (bsc #1025621)

wfeldt:
The driver updates are just not applied. This changes the prepare_rescue script to take care.

This is basically what's described in this blog article:
https://lizards.opensuse.org/2017/02/16/fun-things-to-do-with-driver-updates/


shundhammer:
LGTM

PR #187, opened 2017-06-08
include video drivers from drm-kmp (bsc #1043146)

wfeldt:
The new video drivers are provided as separate kmp for sle12-sp3/leap-42.3.
This patch just supports this setup. The actual decision whether to include
the drm kmp is made in the spec file.


shundhammer:
LGTM

PR #188, opened 2017-06-19
/etc/systemd: keep symlinks instead of files (bsc #1044791)

wfeldt:
systemd added a check to ensure *.service links in /etc/systemd are really
symlinks. So we need to be a bit more careful when copying /etc.

In particular, we now keep symlinks to files except for config files (*.conf)
but resolve links to directories.

### DIFF
--- a/data/initrd/scripts/prepare_rescue
+++ b/data/initrd/scripts/prepare_rescue
@@ -1,11 +1,36 @@
 #! /bin/sh
 
-# ensure some trees are completely writable
+# Ensure some directory trees are completely writable.
+#
+# ... but /etc/systemd will be treated specially (see below): move it aside
+# for now.
+#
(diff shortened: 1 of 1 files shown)


data/initrd/scripts/prepare_rescue
```diff
@@ -1,12 +1,37 @@
 #! /bin/sh
 
-# ensure some trees are completely writable
+# Ensure some directory trees are completely writable.
+#
+# ... but /etc/systemd will be treated specially (see below): move it aside
+# for now.
+#
```
shundhammer: I know it's somewhat paranoid, but what if /etc_systemd already exists? Maybe add

    rm -rf /etc/systemd

to make sure it doesn't.

shundhammer:
LGTM

PR #191, opened 2017-07-24
always load dm_multipath module

wfeldt:
It is required for multipath detection anyway, so we might as well load it
unconditionally during startup.


ancorgs:
LGTM

PR #192, opened 2017-08-04
Fix the default cursor theme (bsc#1051664)

lslezak:
- Fixes https://bugzilla.suse.com/show_bug.cgi?id=1051664
- The icon theme path has been changed, the `dmz-icon-theme-cursors` package now contains two themes/directories: `DMZ-White` and `DMZ-Black`.
- See https://build.opensuse.org/package/view_file/openSUSE:Factory/dmz-icon-theme-cursors/dmz-icon-theme-cursors.spec?expand=1

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -562,7 +562,7 @@ arabic-kacst-fonts:
   /usr/share/fonts/truetype/KacstBook.ttf
 
 dmz-icon-theme-cursors:
-  /usr/share/icons/DMZ
+  /usr/share/icons/DMZ-White
 
 multipath-tools:
   /sbin


kobliha:
One more reason why we should `Always check what we write, whether it makes sense (e.g. directory or file or service exists or so)`

wfeldt:
lgtm

kobliha:
%files
%defattr (-, root, root)
%{_datadir}/icons/DMZ-Black
%{_datadir}/icons/DMZ-White

And these actually exist, so it's really just about one check.

kobliha:
Let's continue on yast-devel (see my e-mail)

PR #193, opened 2017-08-15
drop obsolete xorg drivers (bsc #1053806)

wfeldt:


data/root/root.file_list
```diff
@@ -448,12 +448,7 @@ if arch ne 's390' && arch ne 's390x'
   ?xf86-input-vmmouse:
   xf86-input-void:
   xf86-input-wacom:
-  xf86-video-ati: nodeps
-  xf86-video-cirrus:
```
lslezak: Isn't Cirrus emulation used at some virtualization environment?

lslezak: https://linux.die.net/man/1/qemu-kvm mentions "Cirrus CLGD 5446 PCI VGA card", will KVM work correctly without that driver?

wfeldt: See the discussion in fate #323161. Besides, the point is moot, the packages are already gone. This is just to make installation-images build again.

lslezak: I see, then LGTM...

PR #202, opened 2017-09-19
allow adding packages by running rpm directly

wfeldt:
For this, a new tag 'direct' has been added. E.g. this line in a *.file_list file:

foobar: direct

would basically run 'rpm -i --nodeps foobar.rpm'.

This is needed to handle rpms with lua scripts as there's no other way to
run these scripts than to use rpm.


shundhammer:
LGTM

PR #203, opened 2017-09-21
system-user-root needs to be treated specially

wfeldt:
The package can only be installed 'as-is' as its prein script is in lua.


lslezak:
Just for the record: the `direct` keyword is explaned [here](https://github.com/openSUSE/installation-images/blob/d319ec6872899e195029c177a088a7017b31699e/doc/files.md#packages).

PR #206, opened 2017-10-05
use the same driver updates for caasp as the dist it is based on

wfeldt:
Thorsten (kukuk) requested that caasp (> 1.0) should use the same dir for
driver updates as the sle release it is based on.


shundhammer:
LGTM (until the name and version number of that product changes again which can happen anytime)

PR #215, opened 2017-11-17
add fixed.ko (bsc #1067827)

wfeldt:


wfeldt:
The same as
https://github.com/openSUSE/installation-images/pull/213
that was somehow lost. :-(

PR #217, opened 2017-11-28
Replace ntp with chrony (FATE#323432)

jreidinger:
depends on https://github.com/yast/yast-ntp-client/pull/99 and one more pending autoyast PR

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -404,9 +404,8 @@ perl:
 gpg2:
   /usr/bin/gpg{,2}
 
-ntp:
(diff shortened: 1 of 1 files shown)


wfeldt:
I would go for including everything. Simply add
```
chrony:
```
This will be easier to maintain and just wastes a few kB for unused config scripts.

wfeldt:
lgtm; please adjust the spec (if needed) in system:install:head **before** merging.

jreidinger:
spec adjusted, merging

PR #221, opened 2017-12-08
create groups needed in inst-sys (bsc #1064822)

wfeldt:
... and some other small package adjustments.

Groups are created in pre-install of system-group-hardware meanwhile.

And some packages have moved around their files a bit.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -328,7 +328,7 @@ nscd:
   d /var/run/nscd
 
 ?virtualbox-guest-tools: nodeps
-  /usr/lib/udev
+  /etc/udev
 
 ?xf86-input-vmmouse: nodeps
   /usr/lib/udev/rules.d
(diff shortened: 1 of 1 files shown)


data/initrd/initrd.file_list
```diff
@@ -435,10 +435,13 @@ openssh: nodeps
   E prein
   d etc/ssh
 
+system-group-hardware:
+  /
+  E prein
```
jreidinger: it is called also in openssh and also dbus-1. So does we need to call it more times?

wfeldt: it's the packages prein script

data/initrd/initrd.file_list
```diff
@@ -450,7 +453,7 @@ nscd:
 
 # we just want the user & group entries
 rpcbind:
-  /var/lib
+  E groupadd -r nobody || true
```
jreidinger: can you comment here *why* we need system group nobody?

wfeldt: no we, rpcbind assumes it exists

data/initrd/initrd.file_list
jreidinger: ah, now I get what in documentation mean rpm script. So makes sense.

data/initrd/initrd.file_list
jreidinger: So, can you please add comment here with this explanation, as usually for other cases running prein is enough.

PR #227, opened 2018-01-22
copy grub2 background conditionally (boo#1076577)

lnussel:
The grub2 background image is optional.
Also get rid of the EFI patch. The properties are now set in the
branding package already.


lnussel:
untested :)

lnussel:
looks good, mergeable from my PoV :-) https://openqa.opensuse.org/tests/589148#step/bootloader_uefi/2

PR #228, opened 2018-01-24
set product and menu title in gfxboot.cfg

wfeldt:
Follow-up to bsc#1077188.


lnussel:
lgtm, what happens remains to be seen in openQA

PR #229, opened 2018-01-26
make cypto lib dependencies more flexible

wfeldt:
The catch is that nobody really _requires_ those *-hmac packages. Except in
the rescue system where dracut-fips requires them - so there we can drop the
explicit dependencies.


DimStar77:
LGTM

PR #230, opened 2018-01-29
set grub2 title font and color again

wfeldt:
Got broken with https://github.com/openSUSE/installation-images/pull/227.


lnussel:
Please don't do that. A fix in the sle branding package was missing and is on it's way:
https://github.com/openSUSE/branding/pull/72

wfeldt:
ok, then this is not needed

PR #231, opened 2018-01-29
mk_image: /var/lib/rpm/alternatives was moved to /var/lib/alternatives

DimStar77:
Handle the new path as well

This is just a 'feel right' while looking for something totally unrelated to fix

### DIFF
--- a/bin/mk_image
+++ b/bin/mk_image
@@ -241,7 +241,7 @@ sub fix_alternatives_link
     unlink "$dir/usr/sbin/alternatives";
     unlink "$dir/var/log/update-alternatives.log";
     system "rm -r $dir/etc/alternatives/";
-    system "rm -r $dir/var/lib/rpm/alternatives/";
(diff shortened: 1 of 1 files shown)


bin/mk_image
```diff
@@ -241,7 +241,7 @@ sub fix_alternatives_link
     unlink "$dir/usr/sbin/alternatives";
     unlink "$dir/var/log/update-alternatives.log";
     system "rm -r $dir/etc/alternatives/";
-    system "rm -r $dir/var/lib/rpm/alternatives/";
+    system "rm -r $dir/var/lib{/rpm,}/alternatives/";
```
wfeldt: One of the paths will be missing and you'll always get an error. `rm -rf` would be better.

DimStar77: I was debating with myself about it - the current one is reporting an error in the log and it did not harm, which is why I did leave it at that...

But I agree - no warning is cleaner; will change to -rf

PR #236, opened 2018-03-09
remove default driver update url (bsc #1084510)

wfeldt:
The URL does no longer exist (and there's no agreed-on location anyway).

### DIFF
--- a/data/boot/theme.file_list
+++ b/data/boot/theme.file_list
@@ -5,7 +5,7 @@ if arch eq 'i386' || arch eq 'x86_64'
   gfxboot-branding-<gfxboot_theme>:
     /
     e cp -a etc/bootsplash/themes/<gfxboot_theme>/cdrom/* loader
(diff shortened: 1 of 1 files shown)


data/boot/theme.file_list
```diff
@@ -5,7 +5,7 @@ if arch eq 'i386' || arch eq 'x86_64'
   gfxboot-branding-<gfxboot_theme>:
     /
     e cp -a etc/bootsplash/themes/<gfxboot_theme>/cdrom/* loader
-    e gfxboot --config-file=loader/gfxboot.cfg --change-config install::dud.url=http://download.opensuse.org/update/openSUSE/driverupdate
+    # e gfxboot --config-file=loader/gfxboot.cfg --change-config install::dud.url=http://download.opensuse.org/update/openSUSE/driverupdate
```
joseivanlopez: Is there any reason to keep it commented?

wfeldt: in case we want to re-add in the future people can see how it's done

joseivanlopez: Well, it is also saved in the git history, but ok, LGTM.

PR #241, opened 2018-03-27
s390x: add zpxe.rexx to dvd (bsc #813454)

wfeldt:


ihno:
Thanks

wfeldt:
Changed patch to keep original name `zpe.rexx`.

PR #250, opened 2018-05-17
use reiserfs-kmp (fate #323394)

wfeldt:


etc/config
```diff
@@ -35,9 +35,9 @@ s390    = kernel-s390
 
 ; extra kmps we need (',' separated list)
 [KMP]
-default	=
-i386	= virtualbox-guest,xen
-x86_64	= virtualbox-guest,xen,drm
+default	= reiserfs
```
jreidinger: so reiserfs is not dropped for SLE15?

wfeldt: It is, see fate.

PR #253, opened 2018-07-02
add bcm43xx-firmware (bsc #1099327)

wfeldt:
With 'nodeps' as the package requires kernel-firmware which we don't want
here.


shundhammer:
LGTM

PR #254, opened 2018-07-03
re-add xf86-input-synaptics (bsc #1099807)

wfeldt:


sndirsch:
Not that I would understand the syntax of these .file_list files, but if "?<package>" means, that the content of this package should be added completely to the image, the pull request looks ok to me ...

wfeldt:
It actually means the package is added if it exists.

PR #267, opened 2018-10-01
adjust to glibc-locale package split

wfeldt:


wfeldt:
Neither; notified by email.

PR #268, opened 2018-10-18
pick better fonts for fbiterm (fate #325746)

wfeldt:
fbiterm's 8x16.pcf is pretty much unreadable. Playing around a bit shows
h16.pcf to be much better suited. As it turns out fbiterm looks for it already.

Even more it prefers fonts in the default system font location
/usr/share/fonts/misc over fonts in its own font directory /usr/share/fbiterm/fonts.

So, including fbiterm fonts has just been a waste of space so far.

This patch adds h16.pcf and removes the fbiterm font directory, saving some space.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -126,7 +126,6 @@ cryptsetup:
 device-mapper:
 dosfstools:
 e2fsprogs:
-fbiterm:
 file:
 fillup:
 findutils:
@@ -529,11 +528,17 @@ icewm-lite:
   /usr/bin/icewm-lite
   s /usr/bin/icewm-lite /usr/bin/icewm
 
+# don't include fonts - it prefers /usr/share/fonts/misc anyway
+fbiterm:
+  /usr/bin/fbiterm
+
 dejavu-fonts:
   /usr/share/fonts/truetype/DejaVuSans*.ttf
   r /usr/share/fonts/truetype/DejaVuSansCondensed*
 
 efont-unicode-bitmap-fonts:
+  # fonts used by fbiterm
+  /usr/share/fonts/misc/h16.pcf.gz
   /usr/share/fonts/misc/b16.pcf.gz
 
 ?google-roboto-fonts:


shundhammer:
LGTM

lslezak:
Hm, does that work fully? I'm just playing with the fonts and it seems the `h16.pcf.gz` does not include the pseudographics charcters (mainly frames and filled boxes).

See this screenshot, the frames are missing. And what's most important the background for the inactive `InputField` is missing, the widget is basically invisible.

### h16.pcf

yast-fbiterm-font-h16-en

### The standard console font (eurlatgr.psfu)

lslezak:
# Playing with Fbiterm Fonts

## Standard Text Console

Let's start with the default, this is the standard VGA console using the `eurlatgr` console font:



## Standard Fbiterm font

If you run the `fbiterm` without any options an ugly default font is used. It's quite difficult to read.



## Using the `koi8x16` font

You can use `-a` option to pass a custom font, using `koi8x16.pcf.gz` results in the same (or very similar) look as the plain text console:

yast-fbiterm-font-koi8x16-en

The non-latin languages still look perfectly, this is Japanese:

yast-fbiterm-font-koi8x16-ja
 
Unfortunately `fbiterm` uses a different font for non-ASCII characters, so if the language uses ASCII and some additional characters with accents or umlauts it will look ugly. E.g. `a` and `á` will look completely differently. This is a Czech example:

yast-fbiterm-font-koi8x16-cs

## Unifying the Fonts

You can set the ASCII font (via `-a`) and the Unicode font (via `-m`) for `fbiterm`. So you can set the same font for both. In this example I used `h16_b.pcf.gz`. In the Czech example both `a` and `á` look same. :+1: Unfortunately the glyphs for frames are missing, the worse part is that the background for *Keyboard Test* widget is missing :worried: :

yast-fbiterm-font-h16_b-cs

More over it does not cover non-ASCII, e.g. Japanese:

yast-fbiterm-font-h16_b-ja

## Using the Unifont-like for ASCII

So I tried using `h16.pcf.gz` which looks the same for ASCII characters as the GNU unifont used for the Unicode characters. Then the `a` and `á` look same, Japanase works, but the frame glyphs are still missing. :worried: 

yast-fbiterm-font-h16


Note: when using the `h16*.pcf.gz` fonts I have seen several random YaST crashes during start up, YaST received SIGHUP signal and terminated :boom: , no idea why... :question: 

---

# Summary

Unfortunately it seems that there is no ideal solution, in each case there is something missing or not working properly. :disappointed:

To make it work properly someone would have to build a font covering all parts (ASCII, graphics glyphs, Unicode) with the same look for the ASCII and extended ASCII. And that something out of the YaST scope...

So it seems we cannot run the `fbiterm` unconditionally but only in some cases. :worried:

lslezak:
JFYI with [this workaround](https://github.com/openSUSE/installation-images/pull/269) the line characters are displayed properly. :+1: 

yast-fbiterm-font-h16-screen-cs

lslezak:
JFYI with [this workaround](https://github.com/openSUSE/installation-images/pull/269) the line characters are displayed properly. :+1: 

yast-fbiterm-font-h16-screen-cs

lslezak:
JFYI with [this workaround](https://github.com/openSUSE/installation-images/pull/269) the line characters are displayed properly. :+1: 

yast-fbiterm-font-h16-screen-cs

PR #269, opened 2018-10-19
don't use 'iterm' terminfo (fate#325746)

wfeldt:
fbiterm sets TERM to 'iterm' but this leads to display problems in the
ncurses-UI of yast when using an Unicode (iso10646) encoded font as basic
text font.

Not sure why this happens as fbiterm otherwise displays Unicode chars just
fine even in this situation. Looks like 'iterm' does not match fbiterm's
capabilities. Swapping it for 'screen' seems to make it work fine.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -53,6 +53,11 @@ TEMPLATE skelcd-control-.*:
   # just a template that does nothing
   d .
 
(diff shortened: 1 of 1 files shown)


lslezak:
Yes, it works! :+1: Thanks!

yast-fbiterm-font-h16-screen-cs

PR #272, opened 2018-11-09
suse-module-tools now have a dependency on binutils

wfeldt:
...which is not wanted. So ignore binutils in more places.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -35,6 +35,7 @@ TEMPLATE:
 
 AUTODEPS:
 
+binutils: ignore
 dbus-1-x11: ignore
 diffutils: ignore
 dirmngr: ignore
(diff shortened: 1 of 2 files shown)


data/initrd/initrd.file_list
```diff
@@ -35,6 +35,7 @@ TEMPLATE:
 
 AUTODEPS:
 
+binutils: ignore
```
lslezak: Um, there are no details, I think it would be nice to add a comment, e.g. why we need to ignore the dependency and possibly why it is safe for us. As I can see we do not mention this for the other packages, but IHMO we should start with it.

It will be hard to review these exceptions in the future if do not know why they have been added. (Digging in the Git history is not a nice solution....)

wfeldt: It's explained in the commit message. Note sure the code gets better adding an explanation to every line.

PR #274, opened 2018-11-14
setctsid is dead, long live setsid (bsc #1109290)

wfeldt:
setsid's '-c' option does the same as setctsid (set controlling terminal).


DimStar77:
:+1:

PR #275, opened 2018-11-19
rescue: change service name of nfsserver to nfs-server

DimStar77:
nfs-server is the canonical upstream service name. For a long time,
openSUSE has been shipping a self-crafted service alias as a drop-in

### DIFF
--- a/data/rescue/rescue-server.file_list
+++ b/data/rescue/rescue-server.file_list
@@ -24,7 +24,7 @@ s ../vsftpd.service usr/lib/systemd/system/multi-user.target.wants
 s ../smb.service usr/lib/systemd/system/multi-user.target.wants
 s ../nmb.service usr/lib/systemd/system/multi-user.target.wants
 s ../rpcbind.service usr/lib/systemd/system/multi-user.target.wants
-E chkconfig nfsserver on
+E chkconfig nfs-server on
 
 # add motd
 x /etc/motd /etc/motd


DimStar77:
This should (untested!) fix the issue coming from https://build.opensuse.org/request/show/644756 - where nfsserver.service (the aliased service) is no longer provided

wfeldt:
Can that go to sle15-sp1 as well or will some `if` be needed?

DimStar77:
This should be no problem for SLE15 (even GA) - the nfs-kernel-server rpm has these systemd services packaged:

```
dimstar@zeus:~/Downloads> rpm -qlp nfs-kernel-server-2.1.1-4.32.x86_64.rpm | grep service
warning: nfs-kernel-server-2.1.1-4.32.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID 39db7c82: NOKEY
/etc/sysconfig/SuSEfirewall2.d/services/nfs-kernel-server
/usr/lib/systemd/system/nfs-mountd.service
/usr/lib/systemd/system/nfs-mountd.service.d
/usr/lib/systemd/system/nfs-mountd.service.d/options.conf
/usr/lib/systemd/system/nfs-server.service
/usr/lib/systemd/system/nfs-server.service.d
/usr/lib/systemd/system/nfs-server.service.d/nfsserver.conf
/usr/lib/systemd/system/nfs-server.service.d/options.conf
/usr/lib/systemd/system/nfsserver.service
```

with nfsserver also being the 'alias' for nfs-server.

nfssrever.service contains:

```
[Unit]
Description=Alias for NFS server
# The systemd alias mechanism (using symlinks) isn't rich enough.
# If you "systemctl enable" an alias, it doesn't enable the
# target.
# This service file creates a sufficiently rich alias for nfs-server
# (which is the canonical upstream name)
# "start", "stop", "restart", "reload" on this will do the same to nfs-server.
# "enable" on this will only enable this service, but when it starts, that
# starts nfs-server, so it is effectively enabled.
# nfs-server.d/nfsserver.conf is part of this service.

Requires= nfs-server.service

[Service]
Type=oneshot
ExecStart=/bin/true
RemainAfterExit=yes
```

wfeldt:
ok, let's try

PR #279, opened 2018-12-03
YaST theme cleanup (related to boo#1109310)

lslezak:
- Do not create the /usr/share/YaST2/theme/current symlink,
  it's not needed anymore, the path exists and is a directory.
- Removed the related settings from the theme config and
  the documentation.


wfeldt:
lgtm

PR #280, opened 2018-12-04
s390x: adjust rexx sample script (bsc #1108089, bsc #1108090)

wfeldt:
Removing the version and keeping just 'sles' to avoid having to adjust the
script in future.

Note: the lines are (have to be) padded with spaces to 80 chars.


shundhammer:
LGTM

PR #283, opened 2018-12-10
support adding packages matching a regexp (fate #325482)

wfeldt:
With 'add_all' it is now possible to add a group of packages. The plan is to
use this with 'skelcd-control-.*' to include all skelcd packages without
having to name them explicitly. The spec file would be responsible to manage
the desired list.

### DIFF
--- a/doc/files.md
+++ b/doc/files.md
@@ -170,6 +170,24 @@ glibc:
 systemd: ignore
 ```
 
+### Including packages matching a regexp
+
+To include a group of packages matching a regexp, use `add_all`:
+
+```
+add_all PACKAGE_REGEXP:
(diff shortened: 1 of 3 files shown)


doc/files.md
```diff
@@ -170,6 +170,24 @@ glibc:
...
+To include a group of packages matching a regexp, use `add_all`:
+
+```
+add_all PACKAGE_REGEXP:
+```
+
+Examples:
+
+```
+skelcd-control-.*:
```
ancorgs: The "add_all" part is missing in the example. Isn't it?

shundhammer:
LGTM

PR #284, opened 2018-12-11
add all available skelcd-control packages to inst-sys (fate #325482)

wfeldt:
Before only skelcd-control package dependencies were used. Now install
the actual package(s).


shundhammer:
LGTM

PR #287, opened 2018-12-18
set norepo option if fallback repos are present (fate#325482)

wfeldt:


wfeldt:
not possible this way

PR #294, opened 2019-02-19
include the full set of kernel modules on ppc (bsc #1123349)

wfeldt:
End the special ppc treatment to limit the initrd size.

> Reviewer hint: view diff without whitespace changes.


shundhammer:
LGTM

PR #295, opened 2019-02-21
make openssl-hmac optional

wfeldt:


DimStar77:
:+1: 

Then the logic to add/remove fips comes in the spec file

PR #299, opened 2019-03-12
bash is really in /usr/bin, do not rely in symlink in /bin

wfeldt:
A symlink is not equivalent to a real file, unfortunately.


DimStar77:
makes sense - do we have to take into account that bash.rpm also uses update-alternatives to handle the /usr/bin/sh -> /usr/bin/bash symlink?

wfeldt:
ATM not; the files come out of the package linked correctly.

PR #300, opened 2019-03-12
adjust EFI boot setup to also include RPi firmware files (jsc:SLE-4394)

wfeldt:
The firmware files are in the root dir of the EFI partition. Apply the same
code simplification also to x86.


shundhammer:
LGTM

PR #303, opened 2019-03-25
Establish MicroOS as seperate media

sysrich:
MicroOS is now going to be a standalone distribution for single-service installations, with Kubic being a derivative of MicroOS focused on containers and kubernetes

This requires different products, installation media, etc.. so, here's the installation-images changes required

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -128,9 +128,19 @@ systemd     = CAASP
 
 [Theme Kubic]
 image       = 350
-release     = openSUSE-Tumbleweed-Kubic
+release     = openSUSE-MicroOS
 skelcd      = openSUSE
 skelcd_ctrl = Kubic
+gfxboot     = openSUSE
+grub2       = openSUSE
+plymouth    = openSUSE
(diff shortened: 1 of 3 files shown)


wfeldt:
Does this work? I expect this should also be mapped to TW for driver updates.

https://github.com/openSUSE/installation-images/blob/master/lib/ReadConfig.pm#L928-L933

sysrich:
Thanks, fixed it now, can you re-review? and to answer " does this work" - how would you recommend testing it?

a make on my own machine fails with "mk_image: failto to change owner of tmp to 0:0 at lib/AddFiles.pm line 721 in data/base/base.file_list line 4." but I assume that's nothing to do with problems introduced here, and far more likely a result of this normally being built in OBS, etc..

I was kinda assuming that throwing submit requests, and trusting your review, OBS, and openQA was the right workflow ;)

wfeldt:
'make' works when run as root (unavoidable).

wfeldt:
lgtm

PR #308, opened 2019-04-25
add /dev/btrfs-control to initrd (bsc#1133368)

wfeldt:
Looks like `/dev/btrfs-control` is one of those devices that have to pre-exist.


shundhammer:
LGTM

PR #312, opened 2019-05-03
follow aaa_base package change

wfeldt:
`/etc/sysctl.conf` used to be created in post-in; now it's directly included.


DimStar77:
:+1:

PR #313, opened 2019-05-20
aarch64: ptp_qoriq.ko is now ptp-qoriq.ko

wfeldt:
The `ptp_qoriq.ko` kernel module has been renamed in Factory. Keep also the old name so we continue to work in SLES.


ggardet:
LGTM. Thanks. :) Old name is also required for Leap.

PR #316, opened 2019-06-03
adjust kernel module config (bsc #1136825)

wfeldt:
### Problem

https://bugzilla.suse.com/show_bug.cgi?id=1136825

Kernel module config has changed in SLE12-SP5.

### Solution

Adjust config to include additional modules.


lkocman:
Looks good to me, the modules raised in 1136825 were addressed

PR #319, opened 2019-06-19
Rename product_short to be Kubic relevant

sysrich:
To basically complete what https://build.opensuse.org/package/rdiff/system:install:head/installation-images?linkrev=base&rev=1230 set out to achieve, this patch should stop the files for Kubic being created in /usr/share/tftpboot-installation/openSUSE-MicroOS-x86_64 and instead be created in /usr/share/tftpboot-installation/openSUSE-Kubic-x86_64


wfeldt:
replaced by https://github.com/openSUSE/installation-images/pull/325

PR #323, opened 2019-07-11
aarch64: fix kernel and initrd location in grub config (bsc#1141038)

wfeldt:
### Problem

https://bugzilla.suse.com/show_bug.cgi?id=1141038

"Boot Linux System" menu entry on aarch64 does not work. It loads kernel/initrd from
the wrong location.

### Solution

Fix kernel/initrd path (for SLE12-SP5).

### See also

The same fix in master branch: https://github.com/openSUSE/installation-images/pull/322


shundhammer:
LGTM

PR #325, opened 2019-07-18
allow explicit setting of product name

wfeldt:
### Problem

openSUSE-Kubic is based on MicroOS and all the config files say so. But still it should show up as 'Kubic'.

### Solution

Allow to set the product name explicitly per theme in `etc/config`.

### Note

A similar override exists in `installation-images.spec`.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -142,6 +142,7 @@ gfxboot           = openSUSE
 grub2             = openSUSE
 plymouth          = openSUSE
 systemd           = MicroOS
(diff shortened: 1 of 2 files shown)


shundhammer:
LGTM

sysrich:
For the record - I reviewed and intended to +1'd this before @shundhammer gave his "LGTM"..then got distracted by another phone call and didn't get around to it till now

thanks @wfeldt for this elegant solution! :)

PR #342, opened 2019-11-07
Include KacstOne font for Persian (boo#1092920)

lnussel:


data/root/lang_fonts
```diff
@@ -1,6 +1,7 @@
 # extra fonts to be loaded, comma separated
 font[ar]="/usr/share/fonts/truetype/KacstBook.ttf"
 font[bn]="/usr/share/fonts/truetype/MuktiNarrow.ttf"
+font[fa]="/usr/share/fonts/truetype/KacstOne.ttf"
```
wfeldt: That's not needed; this is only for fonts not already part of the inst-sys.

Which brings us to: the list needs a cleanup as most fonts from that list have been moved into the inst-sys with sle15. But not as part of this pr.

lnussel:
looks like that didn't help unfortunately :(

wfeldt:
Then maybe the other Arabic font needs to be removed or the font config adjusted.

lnussel:
as it works just fine in an installed system I have the feeling that /etc/fonts/suse-post-user.conf screws it up. It wasn't updated since 2006 so that wouldn't surprise me. Maybe we can try without?

data/root/lang_fonts
lnussel: that list is actually read by libyui-qt

https://github.com/libyui/libyui-qt/blob/f6976fc5690fe6fd16dd8343f5bb3ba02ba5a100/src/YQApplication.cc#L215
https://github.com/libyui/libyui-qt/blob/f6976fc5690fe6fd16dd8343f5bb3ba02ba5a100/src/YQApplication.cc#L249

no idea if that is of any use nowadays

PR #343, opened 2019-11-12
another attempt to fix fonts

lnussel:
follow up and revert of #342

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -573,9 +573,6 @@ lklug-fonts:
 arabic-kacst-fonts:
   /usr/share/fonts/truetype/KacstBook.ttf
 
-arabic-kacstone-fonts:
-  /usr/share/fonts/truetype/KacstOne.ttf
-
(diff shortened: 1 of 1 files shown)


lnussel:
fwiw I couldn't test that yet. maybe need to craft a dud that overrides the file?

wfeldt:
You don't need a dud for that. Just boot with `startshell=1` and remove the file before starting yast.

wfeldt:
After playing with it a bit: I've no idea how to make yast prefer KacstOne over KacstBook for Farsi (other than removing KacstBook).

BTW:

- the yast layout is wrong, the progress window should be on the right side
- if you start yast with Farsi locale (`fa_IR`) the word 'Disk' in the progress window is not translated

lnussel:
arabic-noto-naskh
persian-noto-naskh

lnussel:
can we give this a try?

wfeldt:
I'm already testing.

PR #344, opened 2019-11-15
Added linuxrc reboot_timeout option

teclator:
@okurz is adding some patches in order to set the reboot timeout through linuxrc

See https://bugzilla.suse.com/show_bug.cgi?id=1122493

This PR is just for helping him with the changes needed.

See: https://github.com/yast/yast-yast2/pull/977 & https://github.com/yast/yast-installation/pull/823


okurz:
+1

PR #349, opened 2020-01-09
fix initial sshd key generation

wfeldt:
## Problem

SLE-12-SP5:Update got changes that break `installation-images` building.

1. `sshd-gen-keys-start` does not generate sshd keys
2. `dbus-1` postin script fails

## Solution

1.  a backport of https://github.com/openSUSE/installation-images/pull/340 is needed
2. avoid running `chkstat` in `dbus-1`'s postin script; it's (accidentally) the last command and its exit code is passed on; `chkstat` fails after security hardening because `/proc` is not mounted

I'm aware that 2. is not particularly elegant but it gets the job done. The same issue does not appear in later SUSE releases as there `chkstat` is not the last command.

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -294,7 +294,14 @@ dbus-1:
   # not needed
   r /usr/bin/dbus-launch*
   E prein
+
+  # The chmod commands are an evil hack to avoid running chkstat - which
+  # would fail and pass a non-zero exit code as it happens to be the last
+  # command in dbus-1's postin (see bsc#1160594).
+  #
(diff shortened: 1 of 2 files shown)


data/rescue/rescue.file_list
```diff
@@ -294,7 +294,14 @@ dbus-1:
   # not needed
   r /usr/bin/dbus-launch*
   E prein
+
+  # The chmod commands are an evil hack to avoid running chkstat - which
+  # would fail and pass a non-zero exit code as it happens to be the last
+  # command in dbus-1's postin (see bsc#1160594).
```
jreidinger: :+1: for comment

PR #350, opened 2020-01-13
udevd compat symlink has been removed from udev package

wfeldt:
## Problem

The compat symlink 

`/sbin/udevd -> /usr/lib/systemd/systemd-udevd`

is no longer included in the udev package

## Solution

Add the link explicitly. It's a nice thing to have and we can keep existing scripts.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -575,6 +575,9 @@ include theme.file_list
 
 :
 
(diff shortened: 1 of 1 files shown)


DimStar77:
It sounds like a reasonably simple solution without having to find all the other points that refer to udevd. OTOH, it only moves the legacy cruft from one package (systemd) to another (this repo)

wfeldt:
It's only in the inst-sys initrd.

PR #351, opened 2020-01-14
add /sbin/udevadm compat symlink for the moment (bsc#1160890)

wfeldt:
## Problem

`/sbin/udevadm` compat symlink is gone from udev package.

## Solution

Add it here for the moment until all tools have been adjusted.


wfeldt:
On a second thought, this is maybe not the right way after all.

PR #358, opened 2020-02-11
Added libyui-rest-api extension image (bsc#1163275)

lslezak:
- Added an extension image containing the [LibYUI REST API](https://github.com/libyui/libyui-rest-api) packages
- Use case: make the openQA tests easier to implement (they could build a DUD with the packages, but that's not trivial because you need to carefully match the libyui .so version used in the inst-sys, 
 more over it might be more complicated on architectures like aarch64 or s390)
- Activated by the `extend libyui-rest-api` command in the inst-sys
- Tested manually (using a testing image built with `make iso`)
- https://bugzilla.suse.com/show_bug.cgi?id=1163275
- Related .spec file change: https://build.opensuse.org/request/show/773389


Makefile
```diff
@@ -11,9 +11,9 @@ PREFIX  := installation-images-$(VERSION)
 BUILD_ID := $(shell [ -f .build_id ] || bin/build_id > .build_id ; cat .build_id)
 
 # build initrd+modules+gefrickel after tftp (it needs the sha256 sums over the other images)
-COMMON_TARGETS	     := rescue root root+rescue bind libstoragemgmt gdb mini-iso-rmlist tftp initrd+modules+gefrickel
+COMMON_TARGETS	     := rescue root root+rescue bind libstoragemgmt gdb libyui-rest-api mini-iso-rmlist tftp initrd+modules+gefrickel
 # keep in sync with data/boot/tftp.file_list
```
wfeldt: You need to add the image to `tftp.file_list` else it won't be included in the `tftpboot-installation-xxx` packages.

PR #372, opened 2020-04-29
remove ruby parts that are not needed in installation system

wfeldt:
## Problem

Ruby 2.7 comes with `racc`.

## Solution

Take the opportunity and remove some other gems that are not needed during installation as well.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -291,9 +291,12 @@ ruby:
   # avoid update-alternatives
   r /etc/alternatives
   r /usr/bin/rake*
+  r /usr/bin/{y2,}racc*
   r /usr/bin/rdoc*
   r /usr/bin/ri*
   r /usr/bin/bundle*
+  /usr/lib64/ruby/*/{racc,rdoc}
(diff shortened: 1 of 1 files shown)


data/root/root.file_list
```diff
@@ -291,9 +291,12 @@ ruby:
   # avoid update-alternatives
   r /etc/alternatives
   r /usr/bin/rake*
+  r /usr/bin/{y2,}racc*
   r /usr/bin/rdoc*
   r /usr/bin/ri*
   r /usr/bin/bundle*
+  /usr/lib64/ruby/*/{racc,rdoc}
+  /usr/lib64/ruby/gems/*/gems/{minitest,racc,rake,rdoc,test-unit}-*
```
jreidinger, line 299: btw do you remove also gem cache in "/usr/lib64/ruby/gems/*/cache"? it can save some space.  maybe also "/usr/lib64/ruby/gems/*/doc" makes sense if not already removed

wfeldt: gem cache is removed entirely later in the file

jreidinger: good

PR #377, opened 2020-05-18
support Leap-based MicroOS (bsc#1171585)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1171587

There is (will be) a MicroOS based on Leap. Create the expected driver update location for it.

## Solution

Use the `ID_LIKE` tag to differentiate between both variants.

### DIFF
--- a/lib/ReadConfig.pm
+++ b/lib/ReadConfig.pm
@@ -916,9 +916,20 @@ sub get_version_info
 
   die "*** unsupported product: $dist ***" if $dist !~ /^(casp|caasp|kubic|microos|microosng|leap|sles|sled|tumbleweed( kubic)?)$/;
 
-  my $is_tw = $dist =~ /^(microos|tumbleweed( kubic)?)$/;
+  # Kubic is based on TW
+  # MicroOS can be based on TW or Leap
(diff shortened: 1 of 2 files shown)


lib/ReadConfig.pm
```diff
@@ -916,9 +916,20 @@ sub get_version_info
...
+  # Kubic is based on TW
+  # MicroOS can be based on TW or Leap
+
+  my $is_tw = $dist =~ /^tumbleweed( kubic)?$/;
+
+  if($dist eq "microos") {
+    if($config{ID_LIKE} =~ /tumbleweed/) {
+      $is_tw = 1;
+    }
+    elsif($config{ID_LIKE} =~ /leap/) {
```
jreidinger, line 928: I am not sure how often it changes, but maybe using `i` for ID_LIKE can be more robust for future changes in cases.

wfeldt: All the 'machine-readable' tags I've seen so far are lower case:

https://github.com/openSUSE/installation-images/blob/master/etc/os_sample.txt

PR #379, opened 2020-05-22
ensure there's a writable /var/lib/rpm directory (bsc#1171733)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1171733

`/var/lib/rpm` is a symlink to `/usr/lib/sysimage/rpm` - which is read-only. This leads to problems when zypper tries to write to `/var/lib/rpm`.

## Solution

Make `/var/lib/rpm` a writable directory.

## See also

This is a port of https://github.com/openSUSE/installation-images/pull/378 to SLE15-SP2.


aschnell:
LGTM

PR #388, opened 2020-06-23
Save 1.5 MiB RAM by switching inst_setup to dash

mvidner:
`inst_setup` is a long lived program, a process ancestor of YaST, so it pays off to minimize its footprint: from 3360 KiB to 1788 KiB (-1572 KiB) measured by memsample (https://github.com/yast/yast-installation/pull/864)

### Dependencies

- [x] remove bashisms from `/usr/lib/YaST2/startup/common/network.sh`, sourced from here.
      Part of https://github.com/yast/yast-installation/pull/864 : https://github.com/yast/yast-installation/pull/864/commits/7ef133a43f46461fbf749b7c23dca8b294c84b87 (merged)

### Related: CI check? (→ yes, shellcheck)

I know of two tools to check shell scripts for Bash features or bugs: `shellcheck` and `checkbashisms`, and have used them on this PR.

It could be useful to run them in a CI test to detect accidental reintroduction of a bashism.

- `checkbashisms` is *small* (27KiB) and a bit hacky (Perl + regexes). It does not seem to have a 
way to ignore the harmless ocurrence of ulimit:
   > possible bashism in data/root/etc/inst_setup line 164 (ulimit):
   > ulimit -c unlimited
- `shellcheck` is *thorough*, detecting not only bashisms but also potential bugs around glob expansion and quoting, and having an explanation page for each of its diagnostics ([example](https://www.shellcheck.net/wiki/SC2039)). But it is big (17 MiB) and uses Haskell (could it be a problem with expanding the distro rebuild path?)

### Apply to more files? (→ not now)

This repo has many more shell scripts. We should identify those where it would pay off to switch from bash to dash.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ DESTDIR := images/instsys
 
 export ARCH THEMES DESTDIR INSTSYS_PARTS BOOT_PARTS WITH_FLOPPY BUILD_ID
 
-.PHONY: all dirs base fbase biostest initrd \
+.PHONY: all check dirs base fbase biostest initrd \
 	boot boot-efi root rescue root+rescue gdb libyui-rest-api bind libstoragemgmt clean \
 	boot-themes initrd-themes zenroot tftp install \
 	install-initrd mini-iso-rmlist debuginfo cd1 iso
@@ -172,7 +172,7 @@ rescue-server:
 	theme=$(THEMES) image=rescue-server src=rescue filelist=rescue-server fs=squashfs bin/mk_image
 
 root+rescue: base
-	# the next two lines just clean up old files
+	# the next two 'mk_image' runs just clean up old files
 	image=root+rescue fs=none bin/mk_image
 	image=root+initrd src=root+rescue fs=none filelist=root+rescue bin/mk_image
 	bin/common_tree --dst tmp/root+initrd tmp/initrd tmp/root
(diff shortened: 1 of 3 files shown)


wfeldt:
I don't see a compelling need to switch all scripts. Just those that keep running.

If you want to add a test/check target to Makefile, feel free to do so. The build footprint of installation-images is counted in GBs. I don't think adding Haskell will be noticeable.

mvidner:
I added a `make check` target. It is a bit unusual in being able to run *before* `make all`. That is to  Perhaps we should rename it to `make prepcheck`, analogously to `make installcheck`?

The matching OBS change below needs to be made *after* this PR makes its way to OBS otherwise the build will fail.

```diff
Index: installation-images.spec
===================================================================
--- installation-images.spec    (revision d2e182d7f7730c553a2d44bb83e6ee97)
+++ installation-images.spec    (working copy)
@@ -297,6 +297,7 @@
 BuildRequires:  Mesa
 BuildRequires:  Mesa-libEGL1
 BuildRequires:  Mesa-libGL1
+BuildRequires:  ShellCheck
 BuildRequires:  aaa_base
 BuildRequires:  aaa_base-extras
 BuildRequires:  adaptec-firmware
@@ -722,6 +723,7 @@
 %prep
 %setup -n installation-images-%{version}
 rm -f /usr/lib/build/checks/04-check-filelist
+make check
 
 %build
 echo building product "'%product_name_arch'"
```

lslezak:
> If you want to add a test/check target to Makefile, feel free to do so. The build footprint of installation-images is counted in GBs. I don't think adding Haskell will be noticeable.

Just to clarify: Haskell is compiled, the Haskell compiler (ghc) is needed to build ShellCheck, it's not needed for running it. So adding ShellCheck will not increase the i-i build footprint much, it will just make the build chain longer...

Or we could do it the same way as we use Rubocop in YaST - run the check as a part of CI to ensure the Git sources are OK and then do not care in OBS. In fact, this should be actually preferred way as it cannot break if Rubocop is changed/updated in OBS later... And it does not increase the build chain.

PR #391, opened 2020-07-01
generate hostnqn and hostid for NVMe (bsc#1172853)

aschnell:
Generate the hostnqn and hostid for NVMe in the installation and the rescue system.

For https://bugzilla.suse.com/show_bug.cgi?id=1172853.

Also see https://github.com/yast/yast-installation/pull/867 which implements copying hostnqn and hostid to the target system.

### DIFF
--- a/data/initrd/scripts/prepare_rescue
+++ b/data/initrd/scripts/prepare_rescue
@@ -74,6 +74,10 @@ if [ -d /mounts/initrd/update ] ; then
   done
 fi
 
+# create NVMe config files
+nvme gen-hostnqn > /etc/nvme/hostnqn
+cut -d : -f 3 /etc/nvme/hostnqn > /etc/nvme/hostid
+
 if [ "$startshell" = 1 ] ; then
   mount -t proc proc /proc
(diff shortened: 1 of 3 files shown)


data/initrd/scripts/prepare_rescue
```diff
@@ -74,6 +74,10 @@ if [ -d /mounts/initrd/update ] ; then
   done
 fi
 
+# create NVMe config files
+nvme gen-hostnqn > /etc/nvme/hostnqn
```
wfeldt, line 78: Hm, at this point /proc is not mounted and nvme complains a lot - but it does print some string to stdout.

I'm not sure how relevant /proc is for this as this seems to generate just some random string either way. But it's possibly safer to mount and umount /proc around the nvme call. See a few lines below for startshell.

And I see there's also a nvme-gen-hostnqn tool - which doesn't seem to do anything useful at all. But that's OT.

aschnell: I have added mount and umount for /proc.

PR #396, opened 2020-07-14
Rename openSUSE-MicroOS -> MicroOS

lnussel:


lnussel:
will go with https://build.opensuse.org/request/show/820879

PR #397, opened 2020-07-16
move OBS files used for building to git

wfeldt:
## Problem

When submitting the package, sometimes  the spec file has to be adjusted as well. Doing this in a separate
OBS submit request is not ideal.

## Solution

Manage all package related files needed for OBS here in this git repo.

### DIFF
--- a/obs/check-build.sh
+++ b/obs/check-build.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+# Copyright (c) 2004 SuSE Linux AG Nuernberg, Germany. All rights reserved.
+#
+
+function error {
+  echo
+  echo "error: $1"
+  echo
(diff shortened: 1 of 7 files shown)


obs/check-build.sh
```diff
@@ -0,0 +1,55 @@
...
+mkdir -vp $BUILD_ROOT/looptest || exit 1
+cd $BUILD_ROOT/looptest || exit 1
+pwd
+#
+echo "Testing if this host is ok to build this package"
+
+grep -q loop /proc/devices || error "no loop devices"
+
+case `uname -m` in
+  i?86)
```
shundhammer, line 22: No `x86_64` ?

shundhammer: (answered in IRC)

PR #398, opened 2020-07-22
Update the environment variable reference

mvidner:
As I worked on reducing the disk footprint of the installation system (#399) I tried to understand what `mk_image` does at all and I found that the documentation is totally out of date, so I fixed that. Still lots of questions, we can answer them now or later.


doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
+use this to find out actual usage:
+  grep bin/mk_image Makefile | tr ' ' '\n' | sort -u
+-->
 
-  cache the rpms we use, it is actually a bitmask; the most useful value is 4, which
-  will cache the unpacked rpm in a separate directory
+- alternatives=1
 
-- debug=ignore
+  perform a fix-up of update-alternatives links (because...? so that...?)
```
wfeldt, line 20: It was an attempt to solve the update-alternatives problem; it's not used (except in the zenroot image) - instead
the inks are maintained manually iin the *.file_list files.

doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
 
-  ignore noncritical errors
+- disjunct=WORD (eg. `initrd`, `root`)
 
-- debug=account
+  separate WORD from the rest (in what way? so that...?)
```
wfeldt, line 24: The image will include only files that are not in the image specified with disjunct. It's basically a deduplication
used for inst-sys 'extensions' (like 'bind', 'gdb', ...).

The same thing (but done explicitly by calling the common_tree script) is done in the 'root+rescue' make target to
separate initrd, root, and rescue image into 4 disjuct parts.

doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
 
-  use foo instead of tmp/root
+- tmpdir=*TMPDIR*
 
-## rescue
+  Use TMPDIR instead of tmp/root as the temporary directory.
 
-- keeprescue=1
+  The name is a lie. If it really were a temporary directory then the makefile
+  would not need to care to set it to specific values.
```
wfeldt, line 71: The name is chosen because it's below 'tmp'. Semantically it's more of a 'build' directory, though. Yes.

doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
 
-  don't rebuild everything, just recreate (copy the cached tree and compress) the
-  last image (useful to speed up testing)
+- libdeps=DEPS (a comma separated list)
 
-- filelist=foo
+  run `bin/check_libs` on the DEPS (in `tmp/`)
 
-  build using foo.file_list
+- mode=MODE (`add`, `keep`, default: ``)
```
wfeldt, line 50: mode='add' adds files to the image tmp dir (without deleting it first), mode='keep' keeps all data (does nothing) with the tmp dir data

doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
 
-- imagetype=foo / imagetype=foo.gz / imagetype=none
+  don't check for dangling symlinks
 
-  use foo as filesystem for image (.gz: and compress with gzip);
-  none: don't create an image
+- nostrip=1
 
-- imagename=foo
+  ignored! bug? related to commit c109ec79
```
wfeldt, line 60: nostrip is indeed outdated

doc/configoptions.md
```diff
@@ -5,161 +5,102 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
-
-  use tmp/foo & image/foo as temp dir & final image
-
-- tmpdir=foo
-
-  use foo instead of tmp/rescue
+- dist=8.1, dist=8.1-i386+kde
 
-## base
+  build for this release
```
wfeldt, line 82: `dist` is outdated. At least I don't remember using it for a long time. The code around it very likely does not work in any useful way with current products.

doc/configoptions.md
```diff
@@ -171,43 +112,20 @@ The options here are environment variables used by `mk_image`. They are typicall
...
-
 
 ## boot
 
 - boot=small|hd|large|isolinux
 
-  the boot image type we should create; small is a 1.44MB image, hd for
-  boot CD with hd emulation, isolinux for a CD with 'no emulation'; large is
+  the boot image type we should create; `small` is a 1.44MB image, `hd` for
+  boot CD with hd emulation, `isolinux` for a CD with 'no emulation'; `large` is
```
wfeldt, line 121: This is historical. There's nothing to set anymore.

wfeldt:
Adding to the above: `liveeval`, `lang, `fastboot`, and `with_floppy` are also historical. If these are used in the code it's probably obsolete and not working anyway. These variations haven't been built for quite some time.

doc/configoptions.md
mvidner: OK, so can we say this?

> It is analogic to RPM BuildRoot as it holds the file tree in the unpacked state just before being packaged (see FS) in the IMAGE.

doc/configoptions.md
mvidner: So, the default mode:

1. erases the TMPDIR
2. adds files according to FILELIST
3. packs the TMPDIR with FS into IMAGE

And mode=add skips (1), and mode=keep skips (1) and (2)?

doc/configoptions.md
mvidner: I see, it is still used in `bin/mk_boot` but that is not used anywhere. Delete it?

mvidner:
`liveeval` is mentioned in an `if` condition `data/initrd/theme.file_list` . Remove it too?

doc/configoptions.md
mvidner: incorporated

doc/configoptions.md
mvidner: incorporated

doc/configoptions.md
mvidner: Moved it to the Obsolete section

doc/configoptions.md
wfeldt: ok

doc/configoptions.md
wfeldt: yes

doc/configoptions.md
wfeldt: I think so. We don't make anything 'bootable' anymore.

wfeldt:
> `liveeval` is mentioned in an `if` condition `data/initrd/theme.file_list` . Remove it too?

yes, that can go

doc/configoptions.md
```diff
@@ -5,237 +5,122 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
 
-- keeprescue=1
+  look in the `data`/*SRC* directory.
 
-  don't rebuild everything, just recreate (copy the cached tree and compress) the
-  last image (useful to speed up testing)
+- tmpdir=*TMPDIR*
 
-- filelist=foo
+  Use TMPDIR instead of tmp/root as the temporary directory.
```
wfeldt, line 78: Use tmp/TMPDIR instead of tmp/IMAGE...

doc/configoptions.md
```diff
@@ -5,237 +5,122 @@ in `data/*/*.file_list` (see [file list docu](files.md)).
...
-
-- modules=&lt;number&gt;
-
-  build module disk &lt;number&gt;; Note: module disk #1 has the initrd on it, so you _must_
-  build the initrd explicitly before!
-
-
-## liveeval
-
-- lang=cs_CZ
```
wfeldt, line 238: lang=cs_CZ and the following mention of liveeval should also go

PR #399, opened 2020-07-27
Remove pycache to save space (20:4 MiB pre:post squashfs-ing)

mvidner:
Related to:
- https://bugzilla.suse.com/show_bug.cgi?id=1172139
- https://trello.com/c/wnyocCGR/1961-5-reduce-installer-memory-needs-disk-files

Files in the [`__pycache__`][pyc] directories are compiled versions of the sources, used for faster loading. In the inst-sys, saving space is more important than the small speedup, so we delete them.

[pyc]: https://docs.python.org/3/tutorial/modules.html#compiled-python-files

### Size Before (Tumbleweed of 2020-07-04)

```console
$  wget https://download.opensuse.org/tumbleweed/repo/oss/boot/x86_64/{common,root}
(...)
$ du -hc common root
61M     common
73M     root
133M    total
$ pysize() { unsquashfs -ll $1 | grep __pycache__ | awk '{s += $3} END {print s/1024/1024}'; }
$ pysize common
18.8646
$ pysize root
0.151923
```

### Size After

```console
$ du -hc common root
57M     common
72M     root
129M    total
$ pysize root
0
$ pysize common
0
```

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -378,6 +378,7 @@ R s/ Server//g etc/issue
 # remove files we don't want to show up at all
 r /usr/share/{doc,info,locale,man} /usr/src/packages /usr/share/fillup-templates /var/adm/fillup-templates
 r dev/mapper dev/stderr dev/initctl
+e find . -name __pycache__ -print0 | xargs -0 rm -rf
 
 # remove grub2 *.module files, they're not needed
 # ... and strip *.mod files
--- a/data/root/bind.file_list
+++ b/data/root/bind.file_list
@@ -17,3 +17,4 @@ AUTODEPS:
 
 # remove files we don't want to show up at all
 r /usr/share/{doc,info,locale,man} /usr/src/packages /usr/share/fillup-templates /var/adm/fillup-templates
+e find . -name __pycache__ -print0 | xargs -0 rm -rf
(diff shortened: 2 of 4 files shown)


data/rescue/rescue.file_list
```diff
@@ -378,6 +378,7 @@ R s/ Server//g etc/issue
 # remove files we don't want to show up at all
 r /usr/share/{doc,info,locale,man} /usr/src/packages /usr/share/fillup-templates /var/adm/fillup-templates
 r dev/mapper dev/stderr dev/initctl
+e find . -name __pycache__ -print0 | xargs -0 rm -rf
```
lslezak, line 381: I do not know where the cached files can be located, but do we need to scan the whole tree? How long does it take in OBS? I mean scanning e.g. just `usr` could be a bit faster... But maybe it's irrelevant considering how long the i-i build actually takes....

wfeldt: I don't think it would make much of a difference. `/usr` is the largest blob anyway and the file tree has just been created
and is likely still cached.

mvidner: Luckily the measurement is easily obtained, and the answer is: it takes 1 second or less:
```console
$ grep -A1 pycache /var/tmp/build-root-openSUSE_Factory/.build.log
[  333s] running "find . -name __pycache__ -print0| xargs -0 rm -rf"
[  334s] running "touch etc/sysconfig/kernel"
--
[  475s] running "find . -name __pycache__ -print0| xargs -0 rm -rf"
[  475s] running "find var -type f"
--
[ 1348s] running "find . -name __pycache__ -print0| xargs -0 rm -rf"
[ 1348s] running "touch etc/sysconfig/kernel"
```

mvidner:
@wfeldt Anything else? Can we merge?

PR #411, opened 2020-08-26
make exfat support optional

wfeldt:
## Problem

exFAT is not supported in all products.

## Solution

Make it optional and enable only in Tumbleweed.


shundhammer (approved):
LGTM

PR #412, opened 2020-09-07
add watchdog modules (bsc#1176112)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1176112

Watchdog might kick in and reset machine during installation beacuse it thinks it's hanging (i.e. the watchdog driver never gets loaded).

## Solution

Add kernel watchdog modules. This adds around 140 kB (compressed) to the initrd.

## See also

- same for `master`: https://github.com/openSUSE/installation-images/pull/413


shundhammer:
LGTM

PR #414, opened 2020-09-08
Include common critera system role handler (jsc#SLE-12905)

lnussel:


wfeldt:
Is that package included via package deps or do you BuildRequire it?

lnussel:
I thought about adding a buildrequire to the spec file on sle

wfeldt:
The spec is in this git repo; just wanted to mention this.

lnussel:
I wonder why I keep forgetting this

lnussel:
ouch, that was unintentional, wtf

PR #416, opened 2020-09-09
fix library dependency checking

wfeldt:
## Problem

If a library is a symlink to a different library (`libowcrypt` -> `libcrypt` in this case) the library dependency check did not work correctly.

## Solution

In case a library is a symlink to a lib with a different SONAME it is essential to also include the symlink in the elf object list.

The reason is that the library file name is also relevant for soname matching.


ancorgs (approved):
LGTM

PR #418, opened 2020-09-16
create /dev symlinks udev used to create (bsc#1176610)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1176610

Latest udevd no longer creates some symlinks in `/dev`. For a list see above bug report.

This causes some programs to fail (like wicked, in the reported case).

## Solution

Add these symlinks at initrd build time.


DimStar77 (approved):
LGTM

PR #419, opened 2020-09-17
really create /dev symlinks udevd used to create (bsc#1176610)

wfeldt:
## Problem

https://github.com/openSUSE/installation-images/pull/418 does not work.

## Original problem

- https://bugzilla.suse.com/show_bug.cgi?id=1176610

Latest udevd no longer creates some symlinks in `/dev`. For a list see above bug report.

This causes some programs to fail (like wicked, in the reported case).

## Solution

Create the links right before udevd is started. Doing it earlier is no use as linuxrc mounts devtmpfs just before running the `udev_setup` script. Any earlier adjustments are lost.

### DIFF
--- a/data/base/base.file_list
+++ b/data/base/base.file_list
@@ -41,13 +41,6 @@ c 660 0 6 /dev/loop6
 b 7 7 /dev/loop7
 c 660 0 6 /dev/loop7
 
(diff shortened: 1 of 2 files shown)


DimStar77 (approved):
LGTM

PR #420, opened 2020-09-17
plymouthd.defaults is now in branding package 

fcrozat:
with jsc#SLE-11637, /usr/share/plymouth/plymouthd.defaults has been moved to branding package.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -369,8 +369,8 @@ if exists(plymouth)
   plymouth-scripts: nodeps
   plymouth-plugin-script:
   plymouth-branding-<plymouth_theme>: nodeps
-      m /usr/share/plymouth/plymouthd.defaults /etc/plymouthd/plymouthd.conf
-      R s/^Theme=.*/Theme=tribar/ /etc/plymouth/plymouthd.defaults
+      e cp usr/share/plymouth/plymouthd.defaults etc/plymouthd/plymouthd.conf
+      R s/^Theme=.*/Theme=tribar/ /etc/plymouthd/plymouthd.conf
   ?plymouth-theme-tribar:
 endif
 
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -369,7 +369,8 @@ if exists(plymouth)
   plymouth-scripts: nodeps
   plymouth-plugin-script:
   plymouth-branding-<plymouth_theme>: nodeps
-      R s/^Theme=.*/Theme=tribar/ /usr/share/plymouth/plymouthd.defaults
+      m /usr/share/plymouth/plymouthd.defaults /etc/plymouthd/plymouthd.conf
+      R s/^Theme=.*/Theme=tribar/ /etc/plymouth/plymouthd.defaults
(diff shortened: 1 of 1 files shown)


fcrozat:
Another possibility could be to create /etc/plymouthd/plymouthd.conf with the changes, instead of modifying /usr/share/plymouth/plymouthd.defaults

wfeldt:
This does not work for a number of reasons:

- you mix `/etc/plymouth` and  `/etc/plymouthd`
- you mix `plymouthd.conf` and `plymouthd.default`
- using `m foo bar` does not work, `m` copies from the rpm but you can do `e cp foo bar` (no leading `/` in foo and bar)
- you could alternatively do `if exists(plymouth,usr/share/plymouth/plymouthd.defaults)` to test whether the config file is in the package

I was trying to fix all this but tw-20200914 broke `man`, so I can't read up on the current situation. Not that plymouth has any useful docs.

And, finally, shouldn't that be `/usr/etc` meanwhile? If plymouth is repackaging the config stuff.

wfeldt:
Anyway, is there a Staging project for this? Then I can give this a try.

fcrozat:
> Anyway, is there a Staging project for this? Then I can give this a try.

Staging A on IBS 15-SP3.

fcrozat:
> This does not work for a number of reasons:
> 
>     * you mix `/etc/plymouth` and  `/etc/plymouthd`

my bad. 
> 
>     * you mix `plymouthd.conf` and `plymouthd.default`
No, this part is "normal". This is now upstream is doing it.

>     * using `m foo bar` does not work, `m` copies from the rpm but you can do `e cp foo bar` (no leading `/` in foo and bar)

Ok.
>
>     * you could alternatively do `if exists(plymouth,usr/share/plymouth/plymouthd.defaults)` to test whether the config file is in the package

Might be safer. Which method would you prefer ?

> I was trying to fix all this but tw-20200914 broke `man`, so I can't read up on the current situation. Not that plymouth has any useful docs.

Modified plymouth isn't yet in TW, only in Base:System

> And, finally, shouldn't that be `/usr/etc` meanwhile? If plymouth is repackaging the config stuff.
This work is for SLE 15 SP3 where /usr/etc isn't really used. And plymouth has already the concept of defaults predating /usr/etc.

wfeldt:
> * you mix `plymouthd.conf` and `plymouthd.default`
> No, this part is "normal". This is now upstream is doing it.

You create one file, then patch the other.

fcrozat:
> > * you mix `plymouthd.conf` and `plymouthd.default`
> >   No, this part is "normal". This is now upstream is doing it.
> 
> You create one file, then patch the other.

should be fixed now ;)

wfeldt:
This looks much like the modifications I made in my test, but it's `/etc/plymouth` afaics.

fcrozat:
> This looks much like the modifications I made in my test, but it's `/etc/plymouth` afaics.

In that case, feel free to reject this PR and we use your version.

wfeldt:
Ok, done: https://github.com/openSUSE/installation-images/pull/421. Please give it a review to move on.

PR #425, opened 2020-09-29
Extra yast module for common critiera (boo#1176982)

lnussel:


lnussel:
for https://github.com/yast/system-role-common-criteria/pull/4

joseivanlopez (approved):
LGTM

lslezak:
Um, ideally the installer extensions should not need any changes in i-i, it was designed to be extensible even by 3rd party roles/YaST modules. See the example in https://build.opensuse.org/project/show/YaST:extension.

If that somehow does not work for the CC role we could extend/fix it, personally I'd avoid touching i-i whenever possible.

PR #427, opened 2020-10-05
ensure /proc is mounted in chroot environments (bsc#1176972)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1176972
- https://trello.com/c/FePXGCfz

glibc and others rely on `/proc` to be mounted in chroot environments to function properly.

## Solution

Temporarily mount `/proc` if missing. And also set up `/dev/fd` symlinks to `/proc/self/fd`.

### DIFF
--- a/lib/AddFiles.pm
+++ b/lib/AddFiles.pm
@@ -69,6 +69,8 @@ sub find_missing_packs;
 sub rpm_has_file;
 sub fixup_re;
 sub replace_config_var;
+sub mount_proc_and_stuff;
+sub umount_proc_and_stuff;
 
 my $ignore;
 my $src_line;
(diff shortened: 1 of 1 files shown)


lib/AddFiles.pm
```diff
@@ -998,4 +996,64 @@ sub replace_config_var
   return "<$name>";
 }
 
+
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+# Set up /proc and /dev/fd in directory dir if they are missing as a number
+# of tools rely on these (bsc#1160594, bsc#1176972).
```
jreidinger, line 1002: when doing it why not proc mount also "/sys" I think some tools also use it. Otherwise change looks reasonable.

wfeldt: The main reason here is the secure access to file descriptors via `/proc/self`. `/sys` exposes too much of the host system and it would kind of contradict the use of a chroot environment.

Note that using `/proc` is also not without problems as it already exposes quite a lot and processes might influence the host system in unwanted ways.

It's a compromise already and I would not want to stretch it further.

jreidinger: ok

jreidinger (approved):

PR #429, opened 2020-10-06
remove dependency on insserv-compat (bsc#1176501)

wfeldt:
## Problem

There was still a dependency on the insserv-compat package. But that's obsolete and the package is actually not needed.

## Solution

1. remove insserv-compat
1. one of the remaining insserv uses was in the ia32el package; that was for ia64, which is no longer supported
1. there was a keyboard setup script that used to enable utf8 mode on consoles - no longer needed
1. use systemctl for the remaining insserv calls


jreidinger (approved):
cleaning :+1:

PR #430, opened 2020-10-10
Support 32bit ARM (boo#1177264).

hramrach:
Signed-off-by: Michal Suchanek <msuchanek@suse.de>


wfeldt:
This looks basically ok to me. But I couldn't test it.

So, does it build for armv6/7?

hramrach:
It does build images but they do not include grub. Testing is indeed difficult for installtion-images.

hramrach:
Found a typo in the filellists. Now the images build with all expected parts.

wfeldt (approved):
Then let's go for it.

Issue #432, opened 2020-10-15
grub.cgf templating


hramrach:
The data/boot/grub-{aarch64,armv7l,armv6l}.cfg are identical except for the architecture in file path.

Maybe it would be nice to generate them from one file.

I am not sure what would be good place for plugging such templating in the the build system of installation-images, though.

PR #433, opened 2020-10-15
Use one grub.cfg for all ARM architectures.

hramrach:
Fixes: #432

Does the elsif/endif nesting look right?

There was double endif previously. Either we had one extra or we were missing one.

I will try to build some images with this but it will take a while.


wfeldt:
'if' nesting looks better now.

wfeldt:
closing in favor of the more general https://github.com/openSUSE/installation-images/pull/456

hramrach:
Great, thanks

PR #434, opened 2020-10-20
add framework for perl tests

wfeldt:
## Task

- https://trello.com/c/xbmg6qoY

Add a test framework to installation-images and start adding tests.


mvidner (commented):
Thanks!

It's a good test, here's its output (with the verbosity flag that I've added)
```console
$ ./run_tests
test/10_compress_image.t .. ok   
All tests successful.
Files=1, Tests=6,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.04 cusr  0.00 csys =  0.05 CPU)
Result: PASS
$ ./run_tests -v
test/10_compress_image.t .. 
# checking gzip compression
compressing /tmp/gOvOb4xcHJ...
ok 1 - image using gzip was created
ok 2 - is gzip compressed
# checking xz compression
compressing /tmp/gOvOb4xcHJ...
ok 3 - image using xz was created
ok 4 - is xz compressed
# checking default compression
compressing /tmp/gOvOb4xcHJ...
ok 5 - image using default compression was created
ok 6 - is gzip compressed
1..6
ok
All tests successful.
Files=1, Tests=6,  0 wallclock secs ( 0.02 usr  0.00 sys +  0.03 cusr  0.01 csys =  0.06 CPU)
Result: PASS
```

which gives me very good hints for improving the documentation of the module, which is currently too terse:

```
# Usage:
#
#   use CompressImage
#
#   exported functions:
#     CompressImage();
```

As we write the tests it is natural to improve the API docs too.

For example, File::Spec (which is used in i-i) has this [source](https://github.com/Perl/perl5/blob/blead/dist/PathTools/lib/File/Spec.pm) and this [rendering](https://metacpan.org/pod/File::Spec).

Without the pretty formatting, my take at the docs is:

```
# CompressImage($file_name)
# CompressImage($file_name, $method)
# Compress $file_name in place. $method is "gzip" (default) or "xz"
```

Makefile
```diff
@@ -57,7 +57,7 @@ DESTDIR := images/instsys
 
 export ARCH THEMES DESTDIR INSTSYS_PARTS BOOT_PARTS WITH_FLOPPY BUILD_ID
 
-.PHONY: all check dirs base fbase biostest initrd \
+.PHONY: all test check dirs base fbase biostest initrd \
```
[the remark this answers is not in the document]
wfeldt: My plan was to move that within the test framework, along with a check for that dash bug.

run_tests
```diff
@@ -0,0 +1,16 @@
+#! /usr/bin/perl
+
+use TAP::Harness;
+
+$ENV{PERL_TEST_HARNESS_DUMP_TAP} = "test_results";
+
+my $args = {
+  verbosity => 0,
```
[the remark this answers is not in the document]
wfeldt: Sure, can be extended. 'use Getopt::Long' and 'GetOptions()' would be a way. But let's write some more tests first.

The verbose output is also available in the 'test_results' directory.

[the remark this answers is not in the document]
mvidner: Ah, I should have looked. Updated with just a comment.

test/11_lib_conv2_image.t
```diff
@@ -0,0 +1,52 @@
...
+use File::Temp;
+use Test::More;
+
+use vars '$Script';
+$Script = "test";
+
+use MakeExt2Image;
+use Conv2Image;
+
+die("you must be root to run this test") if $>;
```
mvidner, line 13: https://metacpan.org/pod/Test::More suggests
```suggestion
plan skip_all => "you must be root to run this test" if $>;
```

wfeldt: Yes but that doesn't make the tests fail. I wanted it to fail to make a point.

mvidner:
`obs/installation-images.spec` should call `make test`. I thought I did that when adding `make check` but it seems to have got lost.

lib/Conv2Image.pm
```diff
@@ -19,6 +19,7 @@ require Exporter;
 
 use strict 'vars';
 use integer;
+use File::Temp;
 
 sub Conv2Image
```
mvidner, line 24: Like I said about CompressImage, but here it's more important:

Since we have to think about what exactly the `sub` does to get it tested, now is **the** time to write it down in the form of its API docs.
As a user of a library, I must not be forced to read the bodies of subs. Please write doc comments, ideally in a standard format understood by Perl docs tools.

wfeldt: I plan to but one thing after another; this is still WIP.

test/12_lib_make_ext2_image.t
```diff
@@ -0,0 +1,39 @@
...
+for my $task (@$todo) {
+  note("creating ext2 fs: size = $task->{size}, inodes = $task->{inodes}");
+
+  my @res = MakeExt2Image "$tmp_file", $task->{size}, $task->{inodes};
+
+  # fs size must match
+  is((-s $tmp_file)/1024, $task->{size}, "size correct");
+
+  # free blocks are just an approximation
+  cmp_ok($res[0], ">=", $task->{size} * 0.95, "free space not too small");
```
mvidner, line 29: Since the exact sizes seem to matter ...

lib/MakeExt2Image.pm
```diff
@@ -1,73 +1,51 @@
...
+=head1 SYNOPSIS
 
-C<MakeExt2Image.pm> is a perl module that can be used to create Ext2 file
-systems. It exports the following symbols:
+  use MakeExt2Image;
 
-=over
-
-=item *
+  # create a 5000 kbyte file system with 300 inodes.
```
mvidner, line 11: ... shouldn't we be explicit about KiB (1024 B) vs kB (1000B)?
At least in the newly added docs.

mvidner (approved):
Thank you! It's so nice to see docs added to a long used library.

Just some pedantry about 1000 vs 1024.

wfeldt:
Valid point, let's take care in future.

PR #435, opened 2020-10-29
don't forget .lib*.hmac files (bsc#1178208)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1178208

`libcrypto.so` is not working (needed in `insmod`) because its corresponding `*.hmac` file is not available as it's sitting on a separate filesystem for which an `insmod` is needed to mount it...

## Solution

When separating out the early initrd part, don't forget to move `*.hmac` files along with the libraries.


schubi2 (approved):
LGTM

PR #436, opened 2020-11-05
etc: update module.config to match 5.10

jirislaby:
5.10 moved some `net/phy` drivers to `net/mdio` and `net/pcs`. Adapt to that
change.

Since 5.10 (and 5.9.2 too), `nfs_ssc` is now enabled in our kernels as
`nfsd` always needed it. So move it along to `nfsd`.


wfeldt (approved):
lgtm

PR #439, opened 2020-12-11
Don't replace directory symlinks (boo#1029961)

lnussel:
For the usr merge, the top level directories bin, sbin, lib and lib64
will be replaced by symlinks to their counterparts in /usr. Since not
all packages will be switched over from the start, we need to make sure
tar does not screw up the fs tree. This is consistent with rpm behavior.


lnussel:
@wfeldt ?

wfeldt (approved):
While the change in itself is logical I've got the suspicion the symlinking will break the inst-sys in interesting ways.

PR #440, opened 2020-12-15
add system-group-kvm explicitly

wfeldt:
## Task

The system-group-kvm package is required by udev now but would be auto-added too early for its preinstall script to run.

To avoid this, add it explicitly at a better (later) place.


schubi2 (approved):
LGTM

PR #441, opened 2021-01-13
handle the change of the user-visible name of SUSE MicroOS to SLE Micro

jsrain:


wfeldt:
Some bug/jira reference is needed, else the submission will get auto-rejected later on.

jsrain:
@wfeldt Adjused the commit message to include bsc#

wfeldt (approved):
lgtm

PR #442, opened 2021-01-19
Replace tribar plymouth theme for installation with bgrt (boo#1149070)

hellcp:
An extension of openSUSE/linuxrc#245 and yast/yast-installation#904


wfeldt:
The same as with the linuxrc pull request: a reference to either a bug or jira is needed for auto-tooling. You can append it to the pr title.

wfeldt (approved):
lgtm

PR #443, opened 2021-01-22
revert plymouth branding for SLE to tribar

wfeldt:
## Task

The recent plymouth branding switch is unwanted for SLE. See https://github.com/openSUSE/installation-images/pull/442.


fcrozat:
LGTM

PR #445, opened 2021-01-27
do not run prein script for openssh-server in root image

wfeldt:
## Background

openssh-server runs useradd in its prein script. The script used to ignore errors in the past while now it returns an error code.

## Problem

installation-images runs the prein script when building the `root` image. But that's a mistake as it is not needed here because passwd and group files are in the initrd (and openssh-server's prein is run there).

## Solution

Do not run prein script here.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -653,7 +653,6 @@ else
 endif
     /
     t /etc/sysconfig/ssh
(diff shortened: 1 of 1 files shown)


joseivanlopez (approved):
LGTM

PR #448, opened 2021-01-29
Switch to vim-small (boo#1181633)

lnussel:
Also handle /usr/bin/vi pointing to alternatives

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -181,10 +181,10 @@ xz:
 
 <release_theme>-release: nodeps
 
-vim:
+vim-small:
   /
   # avoid update-alternatives
-  e cd usr/bin ; if [ -f vim-nox11 ] ; then ln -snf vim-nox11 vim ; fi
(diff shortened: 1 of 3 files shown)


lnussel:
Couldn't test this yet. The /usr/bin/vi fill will be needed for current vim submission to Factory

wfeldt:
`vi` is a symlink to `vim`. Why would you change that?

lnussel:
is about to become a symlink to /etc/alternatives/vi

wfeldt:
Ok; do you have some bug number to add (to title)? This will also go to sle15-sp3.

lnussel:
vi symlink is boo#1174564
if that goes to sle maybe we need to split off the vim-small part then. Should work there too though as vim-small is available in sle.

wfeldt:
I've seen vim-small in SLE and the package looks no different than the TW one. I don't think anyone will notice the vim
difference.

wfeldt (approved):
lgtm

wfeldt:
But you still need to change the title to contain a bug number.

lnussel:
Ok, I've opened a separate bug for vim-small. The other one really is just about alternatives usage.

PR #450, opened 2021-02-08
Add support for riscv64

andreas-schwab:


wfeldt:
Basically ok; do you have a bug/jira to add to the title? The branch is shared with sle15-sp3 and it will be blocked thete otherwise.

andreas-schwab:
This is factory stuff only.

wfeldt:
Fair enough, but I don't want to branch yet as sle15-sp3 & tw are too much in parallel atm.

wfeldt (approved):
lgtm

PR #451, opened 2021-02-10
Replace Raleway fonts by Poppins

joseivanlopez:
### Problem

New SLE branding uses *google-poppins-fonts* instead of *Raleway*, see https://github.com/yast/yast-theme/pull/132.

* https://jira.suse.com/browse/SLE-15714

### Solution

Provide expected fonts. Note that *google-poppins-fonts* is already included in SLE-15-SP3 project, see https://build.suse.de/package/show/SUSE:SLE-15-SP3:GA/google-poppins-fonts.

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -438,7 +438,7 @@ BuildRequires:  perl-XML-Simple
 BuildRequires:  perl-solv
 BuildRequires:  pinentry
 BuildRequires:  python3-websockify
-BuildRequires:  raleway-fonts
+BuildRequires:  google-poppins-fonts
 BuildRequires:  samba
 BuildRequires:  snapper
 BuildRequires:  suse-module-tools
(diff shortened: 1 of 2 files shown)


wfeldt:
The jira reference is needed either in the title or first line of the commit message.

Does it also work for Tumbleweed?

joseivanlopez:
> The jira reference is needed either in the title or first line of the commit message.

Ok, I fix it.
 
> Does it also work for Tumbleweed?

I think so. TW does not use *Raleway*, see https://github.com/openSUSE/branding/blob/tumbleweed/yast/installation.qss.

obs/installation-images.spec
```diff
@@ -438,7 +438,7 @@ BuildRequires:  perl-XML-Simple
 BuildRequires:  perl-solv
 BuildRequires:  pinentry
 BuildRequires:  python3-websockify
-BuildRequires:  raleway-fonts
+BuildRequires:  goole-poppins-fonts
```
wfeldt: spelling?

wfeldt (approved):
lgtm

PR #452, opened 2021-02-15
remove some obsolete parts from installation-images (bsc#1182291)

wfeldt:
## Task

  - there were references to `pciutils-ids` left that are not used for quite some time now - remove them

  - the library dependency check code could get confused with absolute symlinks to libraries - rewriting it completely to make the code easier to follow and to make it much faster (goes over the file tree just once)

### DIFF
--- a/bin/check_libs
+++ b/bin/check_libs
@@ -1,154 +1,266 @@
 #! /usr/bin/perl
 
-# check dynamic lib dependecies
+# check dynamic library dependecies
 
+use strict;
 use warnings;
+use File::Find;
+use Getopt::Long;
 
-my $verbose = 0;
-while($ARGV[0] eq '-v') {
-  shift;
-  $verbose++;
-}
+use Data::Dumper;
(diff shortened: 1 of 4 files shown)


bin/check_libs
```diff
@@ -48,7 +48,7 @@ while($dir = shift) {
 
   print "finding ELF objects...\n";
 
-  my $cmd = "cd $dir; find . -type f -o -type l | grep -v modules | xargs file -L";
+  my $cmd = "cd $dir; find . -type f | grep -v modules | xargs file -L";
```
shundhammer, line 51: NP: Since this will no longer find symlinks, the `-L` of the `file` command is now no longer needed, I guess.

shundhammer (approved):
LGTM

bin/check_libs
wfeldt: Might be; but it seems the symlinks are really needed - I'll recheck.

PR #456, opened 2021-02-18
unify all EFI-using grub configurations into a single config file (bsc#1182239)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1182239

aarch64 needs to use shim if available.

## Bonus

Finish the work of https://github.com/openSUSE/installation-images/pull/433 and unify all EFI-using architectures.


ggardet:
@wfeldt  Does it make armv6/armv7 to also require a shim?

wfeldt:
No, only if shim is available - and that's controlled in the spec file.

Vogtinator:
FYI, after https://bugzilla.suse.com/show_bug.cgi?id=1176134, `linuxefi`/`initrdefi` is no longer necessary and `linux`/`initrd` works as well.

wfeldt:
Cool; I wasn't sure it's live yet.

dirkmueller:
one thing I noticed is that this replaces the paths with @arch@ - I don't see where this is set, just wanted to point out that armv6/v7 is one of the very few arches where rpm arch (armv7hl) is different than kernel arch (armv7l). and in the old config files it was hardcoded as armv7l.

dirkmueller:
rest looks good to me.

wfeldt:
It's just the `/boot/ARCH/` path on the media; actually the exact value does not matter, as long as it's consistent.

PR #459, opened 2021-02-23
fix adapting grub.cfg for different architectures (bsc#1182593)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1182593

Architecture dependent parts in `grub.cfg` (XML-like enclosed by `<ARCH>...</ARCH>`) were not handled correctly. Only the first non-matching blob was removed.


jsrain (approved):
LGTM, will check the aarch64 installation-images as soon as they build in devel project

PR #460, opened 2021-02-25
go back to plymouth-theme-tribar (bsc#1182755)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1182755

`plymouth-theme-bgrt` requires `plymouth-plugin-label` which (citing its rpm description): *'provides the ability to render text on graphical boot splashes using pango and cairo'*.

So this adds the whole X and cairo font rendering stack to the initrd, which is roughly about 5 MB compressed. Just to show some text.

**Thank you.**

`tribar` can display text on a text console the normal way.

## Solution

Revert to using `plymouth-theme-tribar` and add comment to sources to warn future generations about this.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -363,6 +363,16 @@ pam:
   /lib*/security/pam_deny.so
 
 if exists(plymouth)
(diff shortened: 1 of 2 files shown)


lnussel:
:+1:

PR #461, opened 2021-02-25
Prepare UsrMerge (boo#1029961)

lnussel:


wfeldt:
will be replaced by https://github.com/openSUSE/installation-images/pull/465

PR #462, opened 2021-02-25
support using zram device as root file system (jsc#PM-2253)

wfeldt:
## Task

Allow moving the root file system into a zram device and/or enable swapping to a zram device.

- https://jira.suse.com/browse/PM-2253
- https://trello.com/c/DawiPzcU

## See also

- related linuxrc change, needed to eable this feature: https://github.com/openSUSE/linuxrc/pull/246
- zram documentation: https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html

## Notes

- this just adds the necessary infrastructure, it does not enable this feature
- the changes do not affect the non-zram workflow
- ext2 is used for the root fs, it  has slightly less meta data and tooling overhead than xfs
- zstd compression is used
- if swap into zram is requested, the zram swap device is removed after yast has enabled a 'real' swap device (after the disk preparation stage)
- for usage instructions, check the referenced linuxrc pull request

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -436,6 +436,9 @@ endif
   /usr/lib*/gconv/UTF-32.so
   /usr/lib*/gconv/gconv-modules*
 
+# needed for zram support
+e2fsprogs:
+  /usr/sbin/mkfs.ext2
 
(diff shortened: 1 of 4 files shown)


shundhammer (approved):
LGTM

PR #463, opened 2021-03-02
Add susepaste to the rescue system (bsc#1182212)

ancorgs:
See [bsc#1182212](https://bugzilla.suse.com/show_bug.cgi?id=1182212) which reads: "_Forum posters are often requested to supply additional information by those trying to help solve their problems. Availability of susepaste would facilitate collection of much of that information from those unfamiliar with shell prompts_".

I hope this would be enough to address that, but @wfeldt will know better.

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -163,6 +163,7 @@ squashfs:
 star:
 strace:
 suse-module-tools:
+susepaste:
 sysconfig:
 sysconfig-netconfig:
 systemd-presets-branding-<systemd_theme>:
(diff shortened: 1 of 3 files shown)


wfeldt:
please also add it to the spec in the `obs` directory

wfeldt:
Also, you added it only to the rescue system, not the regular inst-sys.

ancorgs:
I added it to the rescue system because that's what I understood from the bugzilla report. Asking the reporter about it.

ancorgs:
The reporter said "_IMO it could be just as helpful on inst-sys, but probably not be used as often._". So added.

I also added the missing BuildRequires

wfeldt (approved):
lgtm

PR #468, opened 2021-03-10
disable zram swap only if there's another swap active (bsc#1183276)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1183276

zram swap is disabled after partitioning the disk, assuming there's some other swap space active at that point.

This might not be the case.

## Solution

Add check to ensure zram swap is disabled only if there is another swap device active.

## See also

- original implementation: https://github.com/openSUSE/installation-images/pull/462


shundhammer (approved):
LGTM

PR #472, opened 2021-03-31
handle update-alternative symlinks automatically

wfeldt:
## Problem

Some packages use `update-alternative` but the `update-alternative` package is not included to save space und running postinstall scripts is not possible in all cases anyway.

## Solution

Parse the postinstall scripts for `update-alternative` calls and create the expected symlinks.

The parser is rather simplistic and works only for reasonably formatted shell scripts. But for now at least it handles all cases well.

### DIFF
--- a/data/base/base.file_list
+++ b/data/base/base.file_list
@@ -66,12 +66,10 @@ TEMPLATE:
 
 AUTODEPS:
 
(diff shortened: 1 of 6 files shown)


shundhammer (approved):
LGTM

PR #473, opened 2021-04-01
adjust NVME config initialisation (bsc#1183230)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1183230
- https://trello.com/c/fvglnkEV

`/etc/nvme/hostnqn` does not contain system UUID.

## Solution

Adjust config generation according to updated information from

- https://bugzilla.suse.com/show_bug.cgi?id=1183230#c11
- https://bugzilla.suse.com/show_bug.cgi?id=1183230#c22


shundhammer (approved):
LGTM

PR #474, opened 2021-04-07
remove changelog file

wfeldt:
## Task

There was a `changelog` file as make target. That is not really needed. You can look up the commit history with git if needed.

## Bonus

Clean up `VERSION` as well and remove not really needed `.gitignore`.


shundhammer (approved):
LGTM

PR #475, opened 2021-04-07
remove bind-libs BuildRequires

wfeldt:
## Task

`bind-libs` is obsolete and not used at all. Remove dependency.


shundhammer (approved):
LGTM

PR #478, opened 2021-04-14
Fix grub branding for %{arm}

ggardet:


wfeldt (approved):
lgtm

ggardet:
For Leap 15.3, should I submit it to `sle15-sp3` branch?

wfeldt:
Thats's tricky ground. There's no separate Leap branch so indeed sle15-sp3 would be the way to go.

But sle15-sp3 development is more or less closed. There would have to be an official bug against sle15-sp3 to begin with. And it needs to be rated important enough.

ggardet:
I created https://github.com/openSUSE/installation-images/pull/479 targeting sle15-sp3

PR #480, opened 2021-04-15
build with kernel-default-extra on Leap (bsc#1184413, bsc#1183140)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184413
- https://bugzilla.suse.com/show_bug.cgi?id=1183140

Leap 15.3 uses the kernel config from sle15-sp3. That puts some unsupported but common modules into kernel-default-extra, which is not used to build the installation system.

Change spec file to include  kernel-default-extra when building for Leap.


lkocman (approved):
Looks okay to me, approving.

nilxam (approved):
LGTM

PR #481, opened 2021-04-16
enable multipathd in rescue system (bsc#1184686)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184686
- https://bugzilla.suse.com/show_bug.cgi?id=1175560

The rescue system does not have multipath enabled by default. Which causes problems when there is a LVM or RAID setup on top of that.

## Solution

Enable multipath, but with `find_multipaths smart` - which checks for multipath'ed devices and leaves 'normal' devices alone.

### DIFF
--- a/data/rescue/etc/multipath.conf
+++ b/data/rescue/etc/multipath.conf
@@ -1,0 +1,3 @@
+defaults {
+  find_multipaths smart
+}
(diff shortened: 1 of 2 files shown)


mwilck:
Are you shipping the default multipathd service file? In that case it would be possible to pass "multipath=off" on the kernel command line, which would be good to have as a safety measure in case anything goes wrong.

wfeldt:
Yes; and I've just checked, `multipath=off` works.

PR #486, opened 2021-04-16
build with kernel-default-optional on Leap (bsc#1184413)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/485 to `master`.

## Original problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184413

Leap 15.3 uses the kernel config from sle15-sp3. Modules neither in kernel-default nor kernel-default-extra are in kernel-default-optional.

Change spec file to include also kernel-default-optional when building for Leap.


wfeldt:
replaced by https://github.com/openSUSE/installation-images/pull/491

PR #487, opened 2021-04-19
gefrickel: don't skip non-existing

lnussel:
Some libraries like libdl or libc have a link that points to the actual
binary. If the binary is moved away first, the link would be come stale.
Therefore can't check for -e in the loop.


lnussel:
got broken by #450 before

wfeldt (approved):

PR #489, opened 2021-04-19
allow loading of unsupported modules (bsc#1184413, bsc#1183140)

wfeldt:
## Problem

With https://github.com/openSUSE/installation-images/pull/483 and https://github.com/openSUSE/installation-images/pull/485 kernel-default-extra and kernel-default-optional are used to build the installation system on Leap 15.3.

But modprobe will refuse to load these extra modules unless the `allow_unsupported_modules` config option is set.


shundhammer (approved):
LGTM

PR #490, opened 2021-04-20
re-enable SecureBoot on AARCH64 on SLE Micro

jsrain:
enable building the SLE Micro flavor based on OBS macro

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -93,14 +93,12 @@ ExclusiveArch:  do_not_build
 %endif
 
 %if "%flavor" == "SMO"
+%if 0%{?is_smo}
 # build for both Leap and SLE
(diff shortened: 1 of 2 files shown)


obs/installation-images.spec
```diff
@@ -93,14 +93,12 @@ ExclusiveArch:  do_not_build
 %endif
 
 %if "%flavor" == "SMO"
+%if 0%{?is_smo}
```
wfeldt, line 96: There is no `%is_smo`.

jsrain: It will be defined in SUSE:SLE-15-SP3:Update:Products:MIcroOS51, which is the only project where I want this flavour to build.

wfeldt (approved):

PR #492, opened 2021-04-20
allow loading of unsupported modules (bsc#1184413, bsc#1183140)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/489 to master branch.

## Original problem

With https://github.com/openSUSE/installation-images/pull/480 and https://github.com/openSUSE/installation-images/pull/491 kernel-default-extra and kernel-default-optional are used to build the installation system on Leap 15.3.

But modprobe will refuse to load these extra modules unless the `allow_unsupported_modules` config option is set.

### DIFF
--- a/data/initrd/etc/99-unsupported-modules.conf
+++ b/data/initrd/etc/99-unsupported-modules.conf
@@ -1,0 +1,1 @@
+allow_unsupported_modules 1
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

PR #494, opened 2021-04-20
add udev rules for NVMf autoconnect in the installation system (bsc#1184908)

wfeldt:
## Task

The default udev rules for NVMf autoconnect relies on systemd, which is not available during installation.

Add custom udev rules (provided in https://bugzilla.suse.com/show_bug.cgi?id=1184908#c19).


shundhammer (approved):
LGTM

PR #499, opened 2021-04-23
fix NVMf autoconnect udev rule (bsc#1184908)

wfeldt:
## Task

The udev rule provide in https://github.com/openSUSE/installation-images/pull/494 did not work as expected.

Use the [modified version](https://bugzilla.suse.com/show_bug.cgi?id=1184908#c37).


igaw:
Looks good to me. Thanks!

PR #501, opened 2021-04-26
trigger automatic nvme discovery (bsc#1184908)

wfeldt:
## Problem

FC NVME devices are not added automatically.

### DIFF
--- a/data/initrd/etc/90-nvmf-discovery.rules
+++ b/data/initrd/etc/90-nvmf-discovery.rules
@@ -1,3 +1,6 @@
+ACTION=="add", SUBSYSTEM=="fc", DEVPATH=="/devices/virtual/fc/fc_udev_device", \
+      RUN+="/bin/sh -c 'echo add > /sys/class/fc/fc_udev_device/nvme_discovery'"
+
 ACTION=="change", SUBSYSTEM=="fc", ENV{FC_EVENT}=="nvmediscovery", \
       ENV{NVMEFC_HOST_TRADDR}=="*", ENV{NVMEFC_TRADDR}=="*", \
       RUN+="/usr/sbin/nvme connect-all --transport=fc --host-traddr=$env{NVMEFC_HOST_TRADDR} --traddr=$env{NVMEFC_TRADDR}"


igaw:
Should the trigger not happen after the udev daemon has started?

wfeldt:
No idea; I thought that's some kind of settings. Is it really after udev has seen the devices?

wfeldt:
Shouldn't it better be part of a udev rule, then?

igaw:
Do you mean a rule which matches on /sys/class/fc/fc_udev_device/nvme_discovery and then
triggers the discovery? I think this would make it more robust for sure.

igaw:
Let me try to write a rule and test it quickly

wfeldt:
Yes, something like that.

igaw:
With following rule I was able to trigger the discovery via udevd. Though it seems it get's executed twice. So I am not sure if this approach is good idea. I think the safest approach would be to add the echo after two 'triggers'.

BTW, It's very important to match on  'ACTION=="add"'  and not on "change" because otherwise we will have an loop in the rules as the newly discovered devices also have the same DEVPATH and we would match this very same rule again.

> ACTION=="add", SUBSYSTEM=="fc", DEVPATH=="/devices/virtual/fc/fc_udev_device", \
>       RUN+="/bin/sh -c 'echo add > /sys/class/fc/fc_udev_device/nvme_discovery'"

igaw:
I've tested this with on a installed system with systemd running. It could be that the rule would only be executed once with
in the installation image environment.

wfeldt:
Doing the 'add' only once misses changes that might occur later for whatever reason. For example if the driver is provided by a 3rd party driver update or similar.

igaw:
Stupid me, I have to see it run twice. One trigger is from systemd and the other from the rule.

wfeldt:
Can you give the rule file a name, then I can add it.

igaw:
What about adding it as first entry to the existing 90-nvmf-discovery.rules file?

wfeldt:
Whatever you say: :-)

wfeldt:
Something like this?

igaw:
Yes, let's go with this. As you pointed out, this has the beauty that we are not likely to have a timing issue if the subsystem is added later on. Thanks a lot!

PR #502, opened 2021-04-26
extend 'exists' operator to test for a specfic type

wfeldt:
## Task

Add 3rd TYPE argument to `exists` function: that is, allow `exists(PACKAGE, FILE, TYPE)`, like:

```
if exists(filesystem, /lib, l)
...
endif
```

### DIFF
--- a/doc/files.md
+++ b/doc/files.md
@@ -79,7 +79,8 @@ variables don't have a starting `'$'` and are implicitly environment
 variables. The only exceptions to this are `abuild` and `arch`.
 
 Also, you can use `exists(PACKAGE)` to test for a specific package or
-`exists(PACKAGE, FILE)` to test for a file in a package.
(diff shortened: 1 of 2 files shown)


data/initrd/modules.file_list
```diff
@@ -5,7 +5,7 @@ s lib/modules/<kernel_ver>/initrd modules
 <kernel_rpm>:
 
   # temporary
-  if exists(compat-usrmerge-tools)
+  if exists(compat-usrmerge-tools, /lib)
```
lnussel, line 8: should be `filesystem` here I suppose

wfeldt: that's a later pr

lnussel:
the code looks ok, the config change not but needs further adaptions anyway, so ok

PR #503, opened 2021-04-26
make usrmerge case depend on /lib link in filesystem package

wfeldt:
## Task

If `filesystem` package contains a `/lib` link, assume the /usr merge has happened.


lnussel:
:+1:

PR #507, opened 2021-04-28
trigger automatic nvme discovery in udev start script (bsc#1184908)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184908#c56

https://github.com/openSUSE/installation-images/pull/501 didn't work.

Try this way.


igaw:
Looks good to me. Also this time I tested with the installer and the discovery worked and the nvme disk were
available in the disk partitioner.

PR #508, opened 2021-04-29
create NVMe config files before udevd is started (bsc#1184908)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184908#c78

The nvme config is created too late for fc nvme discovery to work correctly.

## Solution

Move the nvme config creation into the `udev_setup` script.

### DIFF
--- a/data/initrd/scripts/udev_setup
+++ b/data/initrd/scripts/udev_setup
@@ -26,6 +26,12 @@ if [ -n "$linuxrc_no_auto_assembly" ] ; then
   echo 'ENV{ANACONDA}="yes"' > /run/udev/rules.d/00-inhibit.rules
 fi
 
+# create NVMe config files before udevd is started (bsc#1184908)
+if [ ! -f /etc/nvme/hostnqn -a -x /usr/sbin/nvme ] ; then
+  { /usr/sbin/nvme-gen-hostnqn || /usr/sbin/nvme gen-hostnqn ; } > /etc/nvme/hostnqn
+  cut -d : -f 3 /etc/nvme/hostnqn > /etc/nvme/hostid
+fi
+
 # start udevd
 echo -n "Starting udevd "
 if [ -n "$linuxrc_debug" ] ; then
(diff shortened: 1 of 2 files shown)


data/initrd/scripts/udev_setup
```diff
@@ -26,6 +26,12 @@ if [ -n "$linuxrc_no_auto_assembly" ] ; then
   echo 'ENV{ANACONDA}="yes"' > /run/udev/rules.d/00-inhibit.rules
 fi
 
+# create NVMe config files before udevd is started (bsc#1184908)
+if [ -x /usr/sbin/nvme ] ; then
```
igaw, line 30: While at it, could we also test if the hostnqn file already exists? If it does, just skip this step? 
This would allow to inject a hostnqn via the command line, e.g. 

`  echo nqn.2014-08.org.nvmexpress:uuid:1a9e23dd-466e-45ca-9f43-a29aaf47cb21 > /etc/nvme/hostnqn
`

igaw (changes requested):

data/initrd/scripts/udev_setup
wfeldt: Command line is actually parsed later. For this to work properly we'd really need a separate linuxrc option.

igaw: Understood. I've modified the line to 

`if [ ! -f /etc/nvme/hostnqn ] && [ -x /usr/sbin/nvme ] ; then`

and added the hostnqn file to initrd via mksusecd. My test setup with qemu doesn't have
have proper interfaces to generate the correct system uuid. Anyway with this I was
able to get it working.

wfeldt:
Ok, added the test for an existing hostnqn file. Note that this also skips the hostid generation.

igaw:
> Ok, added the test for an existing hostnqn file. Note that this also skips the hostid generation.

Should be okay, as if the user adds the hostnqn we can assume the hostid file also exists. I think this a fair assumption.

igaw:
Alright this looks good to me. I really hope we got it finally correct.

wfeldt:
As we're not in a hurry atm, I'd like to wait for feedback from Ricky before merging.

PR #509, opened 2021-05-04
trigger automatic nvme discovery in udev start script (bsc#1184908)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/507 and https://github.com/openSUSE/installation-images/pull/508 to `master` branch.

## Original problems

https://bugzilla.suse.com/show_bug.cgi?id=1184908#c78

- FC NVME devices are not added automatically.
- The nvme config is created too late for fc nvme discovery to work correctly.


shundhammer (approved):
LGTM

PR #510, opened 2021-05-11
openssh-server now uses /usr/etc for its config files (bsc#1185709)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1185709

The latest `openssh-server` package moved its config files to `/usr/etc`. Adjust to work with old and new location.

## Solution

Avoid all the hassle and put a config blob into `/etc/ssh/sshd_config.d`. That works also for the 'old' `openssh-server` package.

## Bonus

Drop support for the ancient unsplit `openssh` package.

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -342,18 +342,13 @@ nfs-client:
 
 ?openssh-fips:
 
-if exists(openssh-server)
-  openssh:
-  openssh-server:
-else
-  openssh:
-endif
-    /
-    E prein
-    E postin
-    # enable root login bsc#1118114
-    R s/^\s*#\s*(PermitRootLogin)\b.*/$1 yes/ /etc/ssh/sshd_config
-
+openssh:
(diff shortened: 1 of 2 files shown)


DimStar77:
> On a related note: why isn't `sshd_config.d` mentioned in our man pages? Compare to the Debian variant.

the sshd_config.d is actually not supported based on 'openssh code', but it's part of the default configuration. Debatable if the man page should talk about how config is changed (the sshd_config has the comments in it)

DimStar77 (approved):

wfeldt:
I was basically expecting that as justification. But with the same reasoning you could point people to read the source code.

The real question is if `sshd_config.d` is going to stay or not. And if it *is* and people can *rely* on it being parsed, then it should be documented. It doesn't really matter if it is parsed because of some C code blob or some config include statement.

Again, see the Debian page as an example. It very prominently explains all this.

Anyway, this is way off-topic.

PR #513, opened 2021-06-01
Keep /usr/lib/ld*so* also if lib64 exists

Vogtinator:
While on x86_64, the runtime linker is /lib64/ld-linux-x86-64.so.2, it's still
in /lib/ on aarch64 (/lib/ld-linux-aarch64.so.1). Without this, the file is
missing and execution fails.

Verified to fix the aarch64 install initrd. Not tested anywhere else.

(CC @lnussel)


lnussel:
:+1:

wfeldt (approved):

PR #517, opened 2021-08-03
fix getting the kernel file name

wfeldt:
## Task

The script needs to identify the actual kernel file in the kernel package. That's not obvious as there are different name variants on each architecture.

The base name is pre-configured in [etc/config](https://github.com/openSUSE/installation-images/blob/sw_108/etc/config#L11-L28). The full name should look something like `/boot/BASENAME-VERSION-VARIANT`.

The existing code assumed there might be name variants without `-VERSION` attached - but this is no longer the case. So simplify the code to avoid ambiguities.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -14,7 +14,7 @@ repository = standard
 ; *** These are actually perl regexps! ***
 ;
 [KernelImage]
-default	= vmlinuz-
+default	= vmlinuz
 alpha	= vmlinuz
 ppc	= vmlinux
 ppc64	= vmlinux
--- a/lib/ReadConfig.pm
+++ b/lib/ReadConfig.pm
@@ -566,7 +566,9 @@ sub KernelImg
     next unless s#.*/boot/##;
     next if /autoconf|config|shipped|version/;		# skip obvious garbage
     my ($f, $l) = split(/ /);
-    if($f =~ m#^$ConfigData{kernel_img}#) {
+    # Explicitly require the kernel file to have a version number attached
+    # with a '-', like NAME-VERSION-VARIANT.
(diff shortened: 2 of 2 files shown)


ggardet:
According to the build log failure from https://build.opensuse.org/package/live_build_log/system:install:head/installation-images:Kubic/ARM/aarch64 it looks for `/boot/Image` on aarch64.

```
[  868s] Warning: Can't identify the real kernel image, choosing the first:
[  868s] Image, Image-5.13.6-1-default
[  868s] 
[  868s] kernel_ver: 5.13.6-1-default
[  868s] kernel_module_dir: /lib/modules
[  868s] --- Building for openSUSE MicroOS (openSUSE-Kubic) aarch64 [lib64], theme Kubic
[  868s] --- Kernel: kernel-default (reiserfs-kmp), Image, 5.13.6-1-default
[  868s] --- Source: /.build.binaries
[  868s] --- Driver Updates: /linux/suse/aarch64-tw
[  868s] we need package kernel-default [5.13.6-1.1.aarch64] {prein,postin,preun,postun}
[  868s] adding package kernel-default [5.13.6-1.1.aarch64] {prein,postin,preun,postun}
[  868s] cp: cannot stat '/usr/src/packages/BUILD/installation-images-17.10/tmp/cache//kernel-default/rpm//boot/Image': No such file or directory
[  868s] mk_image: failed to move /boot/Image to /linux
[  868s]  in data/boot/boot.file_list line 39.
[  868s] make: *** [Makefile:157: boot] Error 1
```

So, the `BASENAME` should be ok. 

Let's see if that helps.

lnussel:
:+1:

PR #519, opened 2021-08-25
ensure /usr/share/pci.ids exists (bsc#1189767)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1189767

`/usr/share/pci.ids` is missing. This causes lspci to display only numerical values.

## Solution

The file is created by a perl script during postinstall of newer versions of the `hwdata` package (basically copying a single file).

To avoid this, create a symlink manually if `pci.ids`  is missing.

### DIFF
--- a/data/rescue/rescue.file_list
+++ b/data/rescue/rescue.file_list
@@ -48,6 +48,12 @@ TEMPLATE wicked|lvm2|syslog-service|util-linux|mdadm:
   /
   E postin
 
+# ensure /usr/share/pci.ids exists
+TEMPLATE hwdata:
+  /
+  e [ -e usr/share/pci.ids ] || ln -s pci.ids.d/pci.ids.dist usr/share/pci.ids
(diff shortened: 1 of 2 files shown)


data/rescue/rescue.file_list
```diff
@@ -48,6 +48,11 @@ TEMPLATE wicked|lvm2|syslog-service|util-linux|mdadm:
   /
   E postin
 
+# ensure /usr/share/pci.ids exists
+TEMPLATE hwdata:
+  /
+  e [ -e usr/share/pci.ids ] || ln -s pci.ids.d/pci.ids.dist usr/share/pci.ids
```
mvidner, line 54: could we make it fail if the target doesn't exist either? to make it a bit future proof

mvidner (approved):
LGTM, with an idea

mvidner (approved):
Thanks!

PR #521, opened 2021-09-02
Fix conditions for turning on/off zram (bcs#1187434)

dgdavid:
## Problem

`inst_setup` is trying to turn on/off zram even when zram feature is not enabled. Thus, a weird/useless error messages are shown when booting

> @swapoff: bad usage
> Try 'swapoff --help' for more information.
>  "cannot create /sys/class/zram-control/hot_remove: Directory nonexistent"

* https://trello.com/c/HKqMXVUK
* https://bugzilla.opensuse.org/show_bug.cgi?id=1187434

## Solution

Fix conditions while trying to turn zram on/off

---

<sub>NOTE: that error message comes from [dash](https://www.man7.org/linux/man-pages/man1/dash.1.html) shell.</sub>

### DIFF
--- a/data/root/etc/inst_setup
+++ b/data/root/etc/inst_setup
@@ -192,10 +192,10 @@ zram_swap_on() {
 }
 
 zram_swap_off() {
-  if [ -b $zram_swap_dev ] ; then
-    swapoff $zram_swap_dev
-    echo $zram_dev_index > /sys/class/zram-control/hot_remove
-  fi
+  [ -b "$zram_swap_dev" ] || return
+
+  swapoff $zram_swap_dev
+  echo $zram_dev_index > /sys/class/zram-control/hot_remove
 }
 
 create_zram_swap_disable_hook() {
@@ -211,7 +211,7 @@ create_zram_swap_disable_hook() {
 #! /bin/sh -x
 
 # at least 2 swap devices
-if [ "\$(wc -l < /proc/swaps)" != 2 ] ; then
+if [ -b "$zram_swap_dev" ] && [ "\$(wc -l < /proc/swaps)" != 2 ] ; then
   swapoff $zram_swap_dev
   echo $zram_dev_index > /sys/class/zram-control/hot_remove
 fi
--- a/data/root/etc/inst_setup
+++ b/data/root/etc/inst_setup
@@ -181,7 +181,7 @@ zram_swap_on() {
     modprobe zram
     zram_dev_index=`cat /sys/class/zram-control/hot_add`
     zram_swap_dev=/dev/zram$zram_dev_index
-    if [ -b $zram_swap_dev ] ; then
+    if [ -b "$zram_swap_dev" ] ; then
       echo zstd > /sys/block/zram$zram_dev_index/comp_algorithm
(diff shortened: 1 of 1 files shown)


wfeldt:
https://github.com/openSUSE/installation-images/blob/63a0765e8f4f5a55c60864b1a86064f30301ed52/data/root/etc/inst_setup#L213-L214

is a bit sloppy, yes. It should be '-ge 2' - but I don't know if dash can do that. The point of this test is to disable zram only if there's some other swap enabled.

Note that `/proc/swaps` includes a header line.

dgdavid:
> The real issue are the lines like
> 
> ```
> if [ -b $zram_swap_dev ]
> ```
> 
> which should be (quoted)
> 
> ```
> if [ -b "$zram_swap_dev" ]
> ```
> 
> else they won't work as expected.

I did a simple test in my running system and the result is the same using and not using quotes. See https://gist.github.com/dgdavid/7124b3d8acf47e3c7f904568d827e0d1

data/root/etc/inst_setup
```diff
@@ -192,10 +192,10 @@ zram_swap_on() {
...
 zram_swap_off() {
-  if [ -b $zram_swap_dev ] ; then
-    swapoff $zram_swap_dev
-    echo $zram_dev_index > /sys/class/zram-control/hot_remove
-  fi
+  [ -b "$zram_swap_dev" ] || return
+
+  swapoff $zram_swap_dev
+  echo $zram_dev_index > /sys/class/zram-control/hot_remove
 }
```
wfeldt, line 199: NP: not sure why you hate `if`

data/root/etc/inst_setup
```diff
@@ -211,7 +211,7 @@ create_zram_swap_disable_hook() {
 #! /bin/sh -x
 
 # at least 2 swap devices
-if [ "\$(wc -l < /proc/swaps)" != 2 ] ; then
+if [ -b "$zram_swap_dev" ] && [ "\$(wc -l < /proc/swaps)" != 2 ] ; then
```
wfeldt, line 214: I think the critical point is

https://github.com/openSUSE/installation-images/blob/master/data/root/etc/inst_setup#L184

which also needs quotes.

dgdavid: I overlooked that one, thanks!

data/root/etc/inst_setup
dgdavid: > NP: not sure why you hate `if`

Me neither :stuck_out_tongue_winking_eye: 

Jokes aside, I prefer guard clauses or early return instead of nested code. That's true. Usually they make the code more readable/understandable (which might be not the case here). Anyway, I was simply trying to avoid a double nesting here because the code was something like 

```bash
if [ -b "$zram_swap_dev" ]; then
  swapoff $zram_swap_dev
  [ -e /sys/class/zram-control/hot_remove] && echo $zram_dev_index > /sys/class/zram-control/hot_remove
fi
```

but then I forgot to _restore_ the `if` clause after getting ride of the not longer needed file existence check.

I'll _fix_ it :)

wfeldt (approved):
lgtm; let's try this :-)

PR #523, opened 2021-09-20
increase minimal ext2 fs size to 128 kiB

wfeldt:
## Problem

Ext2 creation test case fails with latest `e2fsprogs`. The test image (64 kiB) is now considered too small to hold an ext2 fs.

## Solution

Increase minimal ext2 file system size to 128 kiB.

## Note

The test image size is calculated dynamically to fit the test data - which is rather small. So it invariably comes down to the minimal allowed size.


DimStar77 (approved):
LGTM

PR #530, opened 2021-11-03
handle zstd compressed modules

jirislaby:
Modules have `.ko.zst` suffix since kernel 5.15. Handle that in the scripts.


wfeldt:
Thanks for taking care!

But I think you missed one location; I've done https://github.com/openSUSE/installation-images/pull/534 for that.

wfeldt (approved):
lgtm

hramrach:
Can we get this fix on SP4?

PR #534, opened 2021-11-03
handle new .ko.zst kernel module suffix

wfeldt:
## Task

The new kernel (since 5.15) has zstd kompressed kernel modules with `.ko.zst` file name extension.

## See also

- https://github.com/openSUSE/installation-images/pull/530


shundhammer (approved):
LGTM

PR #535, opened 2021-11-10
follow python packages rename: python3-* to python38-*

wfeldt:
## Task

python packages have been renamed to include minor version in package name: `python3-*` to `python38-*`.


DimStar77 (approved):
LGTM

PR #546, opened 2021-11-24
linuxrc handles LIBSTORAGE_* and YAST_* boot options (jsc#SLE-21308)

wfeldt:
## Task

- https://jira.suse.com/browse/SLE-21308

YaST can be controlled via a number of different environment variables (boot options) during installation. They would show up in the bootloader configuration of the target system - which is usually not wanted.

## Solution

Add `LIBSTORAGE_*` and `YAST_*` to the `ptoptions` option.

The net effect is that these options do not end up in the target system's bootloader configuration.

## See also

- related linuxrc change: https://github.com/openSUSE/linuxrc/pull/278


dgdavid (approved):
LGTM

PR #547, opened 2021-11-25
linuxrc handles LIBSTORAGE_* and YAST_* boot options (jsc#SLE-21308)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/546 to SLE15-SP4.

## Original task

- https://jira.suse.com/browse/SLE-21308

YaST can be controlled via a number of different environment variables (boot options) during installation. They would show up in the bootloader configuration of the target system - which is usually not wanted.

## Solution

Add `LIBSTORAGE_*` and `YAST_*` to the `ptoptions` option.

The net effect is that these options do not end up in the target system's bootloader configuration.

## See also

- related linuxrc change: https://github.com/openSUSE/linuxrc/pull/279

### DIFF
--- a/data/initrd/theme.file_list
+++ b/data/initrd/theme.file_list
@@ -70,7 +70,7 @@ e echo "defaultrepo:	`default_repo`" >>linuxrc.config
 
 e echo "KexecReboot:    1" >>linuxrc.config
 
-e echo "PTOptions:	AutoUpgrade,productprofile,addon,XVideo,Screenmode,specialproduct,reboot_timeout" >>linuxrc.config
+e echo "PTOptions:	AutoUpgrade,productprofile,addon,XVideo,Screenmode,specialproduct,reboot_timeout,LIBSTORAGE_*,YAST_*" >>linuxrc.config
 
 if YAST_SELFUPDATE ne ""
   e echo "SelfUpdate:	<YAST_SELFUPDATE>" >>linuxrc.config


data/initrd/theme.file_list
```diff
@@ -70,7 +70,7 @@ e echo "defaultrepo:	`default_repo`" >>linuxrc.config
 
 e echo "KexecReboot:    1" >>linuxrc.config
 
-e echo "PTOptions:	AutoUpgrade,productprofile,addon,XVideo,Screenmode,specialproduct,reboot_timeout" >>linuxrc.config
+e echo "PTOptions:	AutoUpgrade,productprofile,addon,XVideo,Screenmode,specialproduct,reboot_timeout,LIBSTORAGE_*,YAST_*" >>linuxrc.config
```
ancorgs, line 73: Just a question out of curiosity. As far as I understand, these options are case-insensitive. Isn't it? If so, there is any reason to list some of them with CamelCase, others in lower case and others in uppercase?

ancorgs (approved):
LGTM. Especially since this was already approved at #546

data/initrd/theme.file_list
wfeldt: Yes, this is the 'canonical' spelling - the spelling that ends up in install.inf. And YaST does do an exact match.

For example `autoupgrade=1` ends up as `AutoUpgrade: 1` in install.inf.

PR #550, opened 2021-11-26
always include bash -> sh link

wfeldt:
## Task

The bash package does not include the symlink `bash -> sh` anymore. Instead, a new `bash-sh` package exists for that.

To avoid problems, always create this link explicitly. This was done in some places already - now do it in all places where the bash package is used.

### DIFF
--- a/data/base/base.file_list
+++ b/data/base/base.file_list
@@ -70,6 +70,7 @@ AUTODEPS:
 bash:
   /bin/{sh,bash}
   /usr/bin/bash
+  s bash usr/bin/sh
 
 binutils: ignore
 ca-certificates-mozilla:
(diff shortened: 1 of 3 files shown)


data/base/base.file_list
```diff
@@ -70,6 +70,7 @@ AUTODEPS:
 bash:
   /bin/{sh,bash}
   /usr/bin/bash
+  s bash usr/bin/sh
```
DimStar77, line 73: and bin/sh ?

wfeldt: usrmerge forgotten?

DimStar77: no - hence my comment being:

> on UsrMerged systems we're probably ok, not sure if we have to care for the other case still

DimStar77: (might have added that sentence to slow fo you to see it though)

wfeldt: This is only for TW, so usrmerge applies always

DimStar77 (approved):

PR #555, opened 2021-12-17
don't add Y2* install boot options to target system (jsc#SLE-21308)

wfeldt:
## Task

After discussion in review: add also `Y2*` variables to avoid them in target system's boot config.

## See also

- https://github.com/openSUSE/installation-images/pull/556


ancorgs (approved):
LGTM

PR #557, opened 2021-12-21
[package] Ignore missing-PT_GNU_STACK-section rpmlint warnings

DimStar77:
This error shows up on some grub modules on ppc64. As the warning
was non-critical on grub2 build itself, i-i should certainly not
care more for it

### DIFF
--- a/obs/installation-images-rpmlintrc
+++ b/obs/installation-images-rpmlintrc
@@ -1,1 +1,2 @@
-addFilter("arch-independent-package-contains-binary-or-object ")
+addFilter("arch-independent-package-contains-binary-or-object ");
+addFilter("missing-PT_GNU_STACK-section ");


DimStar77:
Imho, it's even debatable if i-i should not have an exception and not do rpmlinting at all; it's a collection of binaries from the distro, that have all been linted already.

DimStar77:
This should allow us to get i-i built on PPC again, eliminating the error seen at
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:PowerPC/installation-images:openSUSE/standard/ppc64le

(the unresolvable regarding gdb in some archs will remain, unfortunately - that's another thing to look at)

wfeldt:
>  it's a collection of binaries from the distro, that have all been linted already.

Well, why are there these `statically-linked-binary` warnings  for ppc grub, then? :-)

Anyway, your change seems currenty not needed. I'll plan to go over the spec in a free minute to clean things up a bit. And I'll also remove the long obsolete installation-images-XXX subpackages along the way.

PR #558, opened 2021-12-21
[packaging] Do not require gdb on ppc ppc64 (big-endian archs)

DimStar77:
gdb is not being built for those archs and is thus not available.

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -368,7 +368,9 @@ BuildRequires:  fbiterm
 BuildRequires:  finger
 BuildRequires:  fonts-config
 BuildRequires:  gamin-server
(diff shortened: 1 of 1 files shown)


DimStar77:
:(

Actually that is incomplete, as it later fails the build with

```
[ 1077s] adding template #1 >.*<
[ 1077s] mk_image: no such package: gdb at lib/ReadConfig.pm line 414, <F> line 4 in data/root/gdb.file_list line 4.
[ 1077s] make: *** [Makefile:190: gdb] Error 2

```

DimStar77:
Oh - it's actually only the test flavor being excluded...

I misread the spec. Thanks for the pointer Andreas!

PR #560, opened 2021-12-31
Add s390x to architecture list for net_repos (boo#1193479)

skriesch:
IBM has been testing our net installations for s390x.
The net iso image is pointing to x86 repos and not to our zsystems repos at the moment.
That is fixing the following bug:
https://bugzilla.opensuse.org/show_bug.cgi?id=1193479

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -55,6 +55,9 @@ ExclusiveArch:  do_not_build
 %ifarch ppc64 ppc64le
 %define the_arch ppc
 %endif
+%ifarch s390x
+%define the_arch zsystems
+%endif
 %ifarch riscv64
 %define the_arch riscv
 %endif
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -74,7 +74,7 @@ ExclusiveArch:  do_not_build
 %define net_repo https://download.opensuse.org/distribution/leap/%{the_version}/repo/oss
 %else
 %define with_exfat 1
-%ifarch %arm aarch64 ppc64 ppc64le riscv64
+%ifarch %arm aarch64 ppc64 ppc64le riscv64 s390x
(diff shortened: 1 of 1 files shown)


obs/installation-images.spec
```diff
@@ -74,7 +74,7 @@ ExclusiveArch:  do_not_build
 %define net_repo https://download.opensuse.org/distribution/leap/%{the_version}/repo/oss
 %else
 %define with_exfat 1
-%ifarch %arm aarch64 ppc64 ppc64le riscv64
+%ifarch %arm aarch64 ppc64 ppc64le riscv64 s390x
```
teclator, line 77: I cannot see s390x listed in https://download.opensuse.org/ports/ and the `the_arch`variable is not set accordingly so I guess it should be also be set to `zsystems` isn't it?

jreidinger: but it is rpm arch, and that is I think correct. E.g. See https://build.opensuse.org/package/view_file/YaST:SLE-15:SP3/yast2-reipl/yast2-reipl.spec?expand=1 ExclusiveArch part

jreidinger: ah, now I get it and agree with you. It needs to set the_arch correctly, otherwise expansion is wrong

jreidinger: @skriesch So please adjust also setting the_arch around line 60

skriesch: Thank you for this qualified review. zsystems is added as the_arch for s390x now. Have a happy new year!

teclator (approved):
LGTM

obs/installation-images.spec
teclator: Thanks for taking care and ...

:fireworks: :tada: Happy New Year :tada: :fireworks:

PR #562, opened 2022-01-19
adjust to recent samba re-packaging

wfeldt:
## Task

samba packages have been re-organized.

I have tested that at least `mount.cifs` still works.


DimStar77 (approved):
LGTM

PR #565, opened 2022-01-20
Add RPi4 arm-trusted-firmware package (bsc#1173489)

hanzyd:


wfeldt:
I don't know what the package does, but adding to BuildRequires does not mean the package ends up on the media.

You might want something similar to

https://github.com/openSUSE/installation-images/blob/master/data/boot/grub2-efi.file_list#L68-L87

maybe?

Also, the `master` branch is only used for Factory. If you need this in SLE15-SP4, there is also a sle15-sp4 branch.

hanzyd:
Yes, I would like this package to be available on installation media. 

arm-trusted-firmware-rpi4 package runs  right before u-boot and provide services for u-boot and linux kernel.

So, besides BuildRequires package should also be added to grub2-efi.file_list, correct?

I will update and resubmit against sle15-sp4 branch.

Thanks!

hanzyd:
This pull request is superseded by PR#566

PR #568, opened 2022-01-24
etc: update module.config to match 5.17

jirislaby:
5.17 brought a new network module mctp-serial (MCTP serial transport).
Add it to the notuseful list as it is very specific module to be built
into installation images.


wfeldt (approved):
Thanks!

PR #569, opened 2022-01-24
re-add mount.nfs (bsc#1195035)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1195035

`mount.nfs` accidentally dropped from initrd as a result of a typo in the usrmerge changes.

## Solution

Add it back and fix a number of warnings and other bit-rot:

1. /usr/sbin/mount.nfs is back
2. remove references left to /bin and /sbin
3. add /usr/sbin/nologin to shut up warnings when adding system users
4. fix warnings about missing group entries
5. remove no longer existing stuff (like bootsplash)
6. fix ruby package handling
7. thai-fonts are in opentype format now
8. fix locale handling (include them in initrd)

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -427,13 +427,11 @@ autoyast2-installation: nodeps
 lvm2: nodeps
   /etc
 
-# note: locale files are added in root image
-if exists(glibc-locale-base)
-  glibc-locale: ignore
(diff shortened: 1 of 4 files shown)


shundhammer (approved):
LGTM.

It's good to see so many things simplified.

PR #570, opened 2022-01-26
use for build proper schema flavor (jsc#SLE-18820)

jreidinger:
Feature: introduce product specific limited set of autoyast modules represented by limited yast2-schema.

Implementation: depending on flavor use proper schema package.


wfeldt:
If both provide yast2-schema, this looks ok.

BTW, if this also for TW, please do a pr for master.

jreidinger:
yes, both provide it. See https://github.com/yast/yast-schema/pull/126/files

PR #571, opened 2022-01-26
use for build proper schema flavor (jsc#SLE-18820)

jreidinger:
Feature: introduce product specific limited set of autoyast modules represented by limited yast2-schema.

Implementation: depending on flavor use proper schema package.

SP4 change https://github.com/openSUSE/installation-images/pull/570


wfeldt (approved):
lgtm

PR #576, opened 2022-02-17
use signed grub.elf on ppc64 (bsc#1196070)

wfeldt:


hramrach:
Note that this will work only on 15.4, earlier releases do not have the signed image.

wfeldt:
This will go only to SP4 and TW.

hramrach:
Thanks for the quick fix

PR #577, opened 2022-02-17
remove obsolete systemd-sysvinit package

wfeldt:


data/rescue/rescue.file_list
```diff
@@ -191,7 +191,7 @@ suse-module-tools:
 sysconfig:
 sysconfig-netconfig:
 systemd-presets-branding-<systemd_theme>:
-systemd-sysvinit:
+systemd-sysvcompat:
```
DimStar77, line 194: Compat should not ever be needed in i-i - the convert only happens on package upgrades from sysv init script to systemd service.

And that era is over too: https://build.opensuse.org/request/show/955694

DimStar77 (changes requested):

PR #579, opened 2022-02-22
include all of s390-tools in initrd (bsc#1195914)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1195914

It's unclear when the s390-tools package will finally include the usrmerge changes.

## Solution

To avoid chosing the wrong path, do not hand-pick individual binaries but include the entire package.


joseivanlopez (approved):
LGTM

PR #580, opened 2022-02-23
fix user/group creation in initrd (bsc#1196331)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1196331

Fix user and group creation in initrd.

There are different ways users and groups are created.

Some are created in packages as needed, some come from dedicated `system-user-xxx` and  `system-group-xxx` packages.

Some packages use preinstall scripts, others RPM LUA commands.


shundhammer (approved):
LGTM

PR #581, opened 2022-03-01
always include bash -> sh links (jsc#SLE-18234)

wfeldt:
## Task

The `bash-sh`  split has reached SLE15-SP4.

This is similar to https://github.com/openSUSE/installation-images/pull/550 but additionally `bin` -> `usr/bin` links have also to be taken care of.


shundhammer (approved):
LGTM

PR #583, opened 2022-03-24
add LeapMicro theme

lkocman:
This is theme as we use it today in LeapMicro 5.2 (original suggestion by Ludwig). We might want to change that over the time but I'd really appreciate getting rid of the patch that we keep as of today.

https://build.opensuse.org/package/view_file/openSUSE:Leap:Micro:5.2/installation-images/leap-micro.diff?expand=1

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -92,6 +92,17 @@ grub2             = openSUSE
 plymouth          = openSUSE
 systemd           = openSUSE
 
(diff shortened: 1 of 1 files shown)


lkocman:
Adding @jsrain to keep him in loop. I was originally surprised that we do not use the openSUSE theme itself. But this is how protoype was put together.

PR #584, opened 2022-03-24
Use new distribution-logos for Leap Micro

lkocman:
* Drop buildrequires on Leap-Micro-release-dvd
  causes unresolvables in Leap Micro 5.2


lkocman:
Our current diff is bit larger, but this would help us to rebase on the latest installation-images and avoid forking it.

wfeldt (approved):
lgtm

PR #585, opened 2022-03-25
Fix creation of openslp user (bsc#1196331, bsc#1197222)

ancorgs:
## Problem

When installing via VNC, the OpenSLP service fails to start making it impossible to find the system via SLP.

That happens because the user `openslp` used to boot the OpenSLP server does not exists in the inst-sys. We are quite sure that's the problem because if the boot process is interrupted using `STARTSHELL` and the user is then manually created with groupadd before proceeding, then OpenSLP runs.

The `openslp` user should be created in the inst-sys by the preinstallation script of the package `openslp-server` that contains the following command:

```
/usr/sbin/useradd -r -g daemon -d /var/lib/empty -s /sbin/nologin -c "openslp daemon" openslp 2>/dev/null || :
```

But that command cannot succeed if the group `daemon` is not there. That dependency, BTW, is declared in the spec file of the package:

```
PreReq:         /usr/sbin/useradd
Requires:       openslp
Requires(pre):  group(daemon)
```

## Solution

Ensure the group `daemon` is there when the prein script of `openslp-server` is executed.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -539,6 +539,8 @@ rpcbind:
 
 # we just want the user & group entries
 openslp-server:
+  # The script assumes the group daemon exists (bsc#1196331, bsc#1197222)
+  E groupadd -r daemon || true
   E prein
 
 # we just want the user & group entries


wfeldt:
To test a SLES build, do (as root):

```
export obsurl=https://api.suse.de obs=SUSE:SLE-15-SP4:GA/standard/x86_64
make THEMES=SLES
```

ancorgs:
I'm having problems to build this locally. I can confirm the change makes `openslp` appear in the /etc/passwd. So this, for sure, fixes an existing problem.

What I cannot confirm is whether that is enough to make OpenSLP run (although I am 99% sure it is) because I have not been able to generate the ISO following the provided instructions.

PR #588, opened 2022-04-21
support leap port of armv7hl (bsc#1198302)

jreidinger:
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1198302

Issue: wrong URL is used for net installer for iso on armv7hl

Solution: Adapt spec file to point to correct url for this architecture.

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -68,7 +68,11 @@ ExclusiveArch:  do_not_build
 %if "%{the_version}" == ""
 %error "bad version string"
 %endif
+%ifarch %arm
+%define net_repo https://download.opensuse.org/ports/%{the_arch}/distribution/leap/%{the_version}/repo/oss
+%else
 %define net_repo https://download.opensuse.org/distribution/leap/%{the_version}/repo/oss
+%endif
 %else
 %define with_exfat 1
 %ifarch %arm aarch64 ppc64 ppc64le


wfeldt:
lgtm

dirkmueller:
LGTM to me as well

obs/installation-images.spec
```diff
@@ -68,7 +68,11 @@ ExclusiveArch:  do_not_build
 %if "%{the_version}" == ""
 %error "bad version string"
 %endif
+%ifarch %arm
+%define net_repo https://download.opensuse.org/ports/%{the_arch}/distribution/leap/%{the_version}/repo/oss
```
ggardet, line 72: I would use `armv7hl` instead of `%{the_arch}` since I suspect `%{the_arch}` value would be `armv7l`

ggardet (changes requested):

obs/installation-images.spec
ggardet: Maybe @dirkmueller could confirm?

ggardet: hmm, it seems it used to work in the past: https://github.com/openSUSE/installation-images/commit/4788feaa282c261b21246fea914671196f8564d0 so I guess It should be ok. Sorry for the noise.

ggardet (approved):

obs/installation-images.spec
jreidinger: np, in general I try to have it more generic if we add more ports, then it should need just to change condition and not url.

PR #594, opened 2022-05-16
adjust to python 3.10

wfeldt:
## Task

python version was upgraded to 3.10. It is used explicitly in a few places; adjust there.


dirkmueller:
no longer needed, #598 superseded this

PR #598, opened 2022-06-07
Adjust to ignore all python 3.x versions

dirkmueller:
* The websockify dependency is not needed as it is required by novnc,
which is already listed.

* Remove python2-* stuff as that is no longer built in tumbleweed

* Ignore all the numpy versions to be futureproof.

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -104,7 +104,6 @@ nfsidmap: ignore
 pam-config: ignore
 permissions: ignore
 pinentry: ignore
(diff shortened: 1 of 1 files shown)


DimStar77:
Closely related to https://github.com/openSUSE/installation-images/pull/594

dirkmueller:
> Closely related to #594

indeed. this version should be mergeable right away as it is hopefully compatible with standard tumbleweed as well as `Staging:A` . just discussed this in person with steffen.

PR #602, opened 2022-07-15
Leap Micro support jsc#SMO-126

lkocman:
* respect recent rename to SLE-Micro


wfeldt (approved):
lgtm

Just one question: shouldn't LeapMicro show up in `_multibuild` at some point?

lkocman:
it should yes, we do actually always remove all other flavors! Let me add it

lkocman:
See #603  for the master branch pr

Issue #607, opened 2022-08-14
shadow: pam files move from /usr/etc to /usr/lib/pam.d


DimStar77:
Shadow has been submitted for inclusion into factory, changing the location of the pam config, following recent changes in pam too

https://build.opensuse.org/request/show/994683

This results in build fails of i-i that need to be addressed
```
[  394s] /usr/sbin/groupadd -r -g 2 daemon
[  394s] groupadd: PAM: Critical error - immediate abort
[  394s] mk_image: execution of prein script failed at lib/AddFiles.pm line 800 in data/initrd/initrd.file_list line 53.
[  394s] make: *** [Makefile:92: initrd] Error 255
[  394s] error: Bad exit status from /var/tmp/rpm-tmp
```

DimStar77:
If I did this right, it should address the build fails seen at
https://build.opensuse.org/package/show/openSUSE:Factory:Staging:O/installation-images:openSUSE

PR #610, opened 2022-08-18
Allow shadow to put pam files to /usr/lib/pam.d

DimStar77:
Fixes issue #607


DimStar77:
If I did this right, it should address the build fails seen at
https://build.opensuse.org/package/show/openSUSE:Factory:Staging:O/installation-images:openSUSE

wfeldt (approved):
lgtm

PR #611, opened 2022-08-22
require glibc-locale explicitly

wfeldt:
## Problem

Due to a libyui dependency change glibc-locale is no longer implicitly added.

## Solution

Add glibc-locale to BuildRequires.


DimStar77 (approved):
lgtm

PR #614, opened 2022-11-07
etc: update module.config to match 6.1

jirislaby:
6.1 brought a new group in networkintg: Power Sourcing Equipment. Add it all to the other list as the modules are very specific modules to be built into installation images.


jirislaby:
Frinedly ping.

wfeldt (approved):

PR #623, opened 2023-02-08
support 32-bit UEFI firmware on x86_64/i386 (bsc#1208003, jsc#PED-2569)

wfeldt:
## Task

- https://jira.suse.com/browse/PED-2569
- https://bugzilla.suse.com/show_bug.cgi?id=1208003

Make installation media  bootable with 32-bit UEFI firmware.

## Notes

This has two effects:

1. The i386 Tumbleweed installation media are now  bootable on 32-bit UEFI systems.
2. The x86_64 Tumbleweed media **would** be bootable on 32-bit UEFI systems once the `grub2-i386-efi` package has been made available on x86_64 (it currently is not).

No change in our tooling chain is needed for this (I think).

The patch does **not** add support for 32-bit UEFI in our installer YaST. The installed system might not be bootable unless you manually adjust it. Running `pbl --install` once should do the trick (assuming you have created an EFI system partition).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,9 @@ COMMON_TARGETS	     := rescue root root+rescue bind libstoragemgmt gdb libyui-re
 COMMON_INSTSYS_PARTS := config rpmlist root common rescue bind libstoragemgmt gdb libyui-rest-api
 
 ifneq ($(filter i386, $(ARCH)),)
-ALL_TARGETS   := initrd-themes initrd boot boot-themes $(COMMON_TARGETS) zenroot
+ALL_TARGETS   := initrd-themes initrd boot-grub2-efi boot boot-themes $(COMMON_TARGETS) zenroot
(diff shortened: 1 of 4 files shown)


rokejulianlockhart:
Hearing that this paves the way somewhat for even x86-64 installers to boot from 32-bit EFIs is significantly more than I expected. Thanks, loads.

wfeldt:
I've created https://bugzilla.suse.com/show_bug.cgi?id=1208057; let's see what happens.

wfeldt:
Ok, both 32 bit and 64 bit booting should work with next (or so) TW.

rokejulianlockhart:
@wfeldt, does that mean that the currently available ISOs include this ability, or not? I don't mind waiting if I can track the generation of new ones at Open Build Service etcetera, but I've no idea how to check. (Thanks.)

PR #647, opened 2023-06-05
usrmerge: Fix build with usrmerged kernel modules (boo#1211796)

hramrach:
With depmod searching for modules in /usr/lib existing installation scripts fail. This requires the updated kmod.


hramrach:
While some parts of the project are unused and need not be updated there are probably places outside the part that buiilds modules.dep that should be updated as well.

Also I am pretty sure changing hardcoded /lib to harddcoded /usr/lib will break installation-images for SLE/Leap.

So while some parts of this are likely superfluous that smaller change looks less than the minimum required.

PR #651, opened 2023-08-01
Drop dependency on Leap-Micro-release-dvd for LeapMicro

lkocman:
Hello,

this is a change that we have been using in our LeapMicro 5.X spec for quite some time, SLEM 5.5 installation images fail on `nothing provides Leap-Micro-release-dvd`

See https://build.opensuse.org/package/view_file/openSUSE:Leap:Micro:5.4/installation-images/installation-images.spec?expand=1


lkocman:
https://bugzilla.suse.com/show_bug.cgi?id=1214688

PR #658, opened 2023-09-13
BuildConflict libpxbackend-1_0-mini (boo#1215290)

DimStar77:
wget pulls in libproxy1 which in turn pulls in libpxbackend-1_0;
to counter cycles, this exists as mini. Declare the mini to be
in conflict for installation-images.


DimStar77:
Build tested at https://build.opensuse.org/package/show/home:dimstar:Factory/installation-images

wfeldt:
I'd rather try and drop `wget`. We have `curl` and wget seems not to be used by yast implicitly, afaics.

PR #667, opened 2023-10-31
Add DTB installation support for Lenovo X13s (bsc#1215647)

chesterlintw:
The X13s DTB needs to be installed on the installer's ESP otherwise the installer won't boot on Lenovo X13s since its UEFI FW does not support ACPI for Linux, and it needs to load the upstream device-tree before running into grub2 and linux kernel.

### DIFF
--- a/data/boot/grub2-efi.file_list
+++ b/data/boot/grub2-efi.file_list
@@ -91,8 +91,14 @@ if arch eq 'aarch64' || arch eq 'armv7l' || arch eq 'armv6l'
     else
       u-boot-rpi3:
         /
+    endif
+    e mv boot/vc/* .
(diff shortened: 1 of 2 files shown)


wfeldt (commented):
Shouldn't `lenovo-x13s-firmware-dt` show up in `obs/installation-images.spec` or is it automatically required by some other package?

chesterlintw:
Yes, you are right. I will revise it. Thanks for the reminder.

chesterlintw:
Hi Steffen,

I have pushed one more commit for adding the BuildRequires directive. Thanks.

wfeldt:
Thanks!

PR #671, opened 2023-11-03
add TPM support (bsc#1216835)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/670 to SLE15-SP5.

## Original task

- https://bugzilla.suse.com/show_bug.cgi?id=1216835

Add TPM support to installation system.

This only makes the TPM accessible. TheTPM  is not used in any way by default.

In detail:

- `tpm2.0-tools` are added
- auto-start `tpm2-abrmd` (the TPM communication daemon) if `/dev/tpm0` exists

With this, you can read the PCR values with `tpm2 pcrread`, for example.

### DIFF
--- a/data/initrd/initrd.file_list
+++ b/data/initrd/initrd.file_list
@@ -102,6 +102,8 @@ sysconfig-netconfig:
 sg3_utils:
 systemd-presets-branding-<systemd_theme>:
 tar:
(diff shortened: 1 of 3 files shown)


wfeldt:
Not needed.

PR #678, opened 2023-12-17
Add dbus-1-daemon

gmbr3:
dbus-1 is being switched to dbus-broker but initrd doesn't have systemd so it can't be used so keep dbus-daemon


gmbr3:
> Is there is Staging project to test the changes?

Staging project F
dbus-1:
https://build.opensuse.org/request/show/1137655

dbus-broker:
https://build.opensuse.org/request/show/1134285

systemd-presets-common-SUSE:
https://build.opensuse.org/request/show/1132887

PR #680, opened 2024-01-09
update name noto-naskharabic-fonts to google-noto-naskharabic-fonts

Pi-Cla:
See https://build.opensuse.org/request/show/1137228#comment-1870055

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -564,7 +564,7 @@ khmeros-fonts:
 
 lklug-fonts:
 
-noto-naskharabic-fonts:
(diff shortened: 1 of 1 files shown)


wfeldt:
What about `noto-sans-fonts`? Will that also be replaced by `google-noto-sans-fonts`?

wfeldt:
You'd also want to adjust obs/installation-images.spec.

wfeldt:
I've created https://github.com/openSUSE/installation-images/pull/684 which included the BuildRequires line.

PR #698, opened 2024-02-15
remove haveged (bsc#1219910)

OttoHollmann:
Starting from Linux kernel v5.4, the HAVEGED inspired algorithm has been included in the Linux kernel. Additionally, since v5.6, as soon as the CRNG gets ready, /dev/random does not block on reads anymore.


wfeldt:
Let's go for it.

Thanks!

PR #702, opened 2024-03-21
Require Mesa-dri

bmwiedemann:
`Mesa-dri` is important for the Qt UI
and so far was pulled in via a runtime Requires in `Mesa` but we want to drop that for cleaner build dependencies for more parallel builds.


Vogtinator:
This would fix https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:M/installation-images:openSUSE/standard/x86_64

Vogtinator:
@wfeldt ping

wfeldt:
I would remove the `Mesa` line and not replace it with `Mesa-dri`.

PR #707, opened 2024-04-27
refresh ssh_host_rsa_key (jsc#ped-129)

olafhering:
sshd rejects old rsa keys with less than 1024 bits in length since version 7.6 (https://www.openssh.com/txt/release-7.6):

    error: main: Host key /etc/ssh/ssh_host_rsa_key: Invalid key length

Refresh the existing key from 2002-08-31 with the following commands:

    rm data/root/etc/ssh/ssh_host_rsa_key*
    /usr/bin/ssh-keygen -A -t rsa -f data/root
    /usr/bin/ssh-keygen -c -f data/root/etc/ssh/ssh_host_rsa_key


olafhering:
This needs to be merged into branch sle15-sp6, due to the openssh version upgrade.

PR #708, opened 2024-05-02
Drop xiterm dependency

hramrach:
fbiterm (xiterm) is a dead project - the latest change is from 2004. It does not appear to work well on recent kernels with several drivers. Factory has mlterm-sdl2 and SDL2 updated to not cache kernel keymap, both are projects actively maintained upstream. Now also fails to build with gcc 14
Related JIRA references: jsc#PED-3655 jsc#PED-7839

### DIFF
--- a/data/root/root.file_list
+++ b/data/root/root.file_list
@@ -520,10 +520,6 @@ libgbm*:
 icewm-lite:
   /usr/bin/icewm-lite
 
-# don't include fonts - it prefers /usr/share/fonts/misc anyway
-fbiterm:
-  /usr/bin/fbiterm
-
 if exists(adobe-sourcesanspro-fonts,/usr/share/fonts/truetype/SourceSansPro-Regular.otf)
   adobe-sourcesanspro-fonts:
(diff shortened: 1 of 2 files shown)


shundhammer:
https://bugzilla.suse.com/show_bug.cgi?id=1224053#c1

> We know very well that it's been a dead project for many years; but it was pretty much the only reliable way to get i18n support for the YaST NCurses UI for languages outside the Latin1 (Western Europe) and parts of the Latin2 (Eastern Europe; CZ, PL, HR) world. In particular, CN, KO, JP were problematic without it.
>
> Dropping fbiterm would pretty certainly mean no more support for Chinese, Japanese, Korean in YaST NCurses; and I am unsure how well cyrillic locales (Russian, Ukrainian, Bulgarian etc.) or Greek would work.

wfeldt:
Replaced by https://github.com/openSUSE/installation-images/pull/715

shundhammer:
See also https://github.com/yast/yast-installation/pull/1117

PR #712, opened 2024-05-07
increase XZ compression level for initrd (bsc#1223982)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/711 to SLE15-SP6.

## Original problem

Initrd size gets too big on ppc64. For a small size reduction, increase initrd compression level.

Using zstd instead of xz gets worse results.


hramrach:
There is also -e option that may or may not be helful.

Using multiple threads makes the image builds non-reproducible.

PR #716, opened 2024-05-14
gefrickel: Do not compress on ppc64le (bsc#1223982 jsc#PED-8374).

hramrach:


hramrach:
The modules.dep file needs to be rebuilt after the modules are uncompressed because it stores the module filename including the extension.

The symptom is that some modules load fine because they have no dependencies besides the kernel (eg. most network and storage drivers) but modules that depend on other modules fail to load (notably sd_mod).

wfeldt:
Replaced by https://github.com/openSUSE/installation-images/pull/727 and https://github.com/openSUSE/installation-images/pull/729.

PR #720, opened 2024-05-31
Use dedicated path for Slowroll tftpboot files

bmwiedemann:
Use dedicated path for `tftpboot-installation-openSUSE-Tumbleweed-Slowroll` files
so in can be installed in parallel to the Tumbleweed files


wfeldt:
No, that's not a good way. this hides the product name in several places.

You can, for example, set the product name explicitly per theme, like this:

https://github.com/openSUSE/installation-images/pull/722

This ends up internally as `$ConfigData{product_short}` which is only used for the tftp directory name.

PR #726, opened 2024-07-22
make some compression settings configurable (bsc#1223982, jsc#PED-8374)

wfeldt:
## Task

- https://jira.suse.com/browse/PED-8374

Introduce environment variable `instsys_no_compression` to allow influencing compression strategy in some places. Notably for ppc64.

## Bonus

Limit compression threads in OBS to 4 threads, for reproducible builds.

### DIFF
--- a/doc/compression.md
+++ b/doc/compression.md
@@ -1,0 +1,13 @@
+# Adjusting compression
+
+The environment variable `instsys_no_compression` accepts a comma-separated
+list of values to omit compression in certain places.
+
+It is usually not a good idea to change the default settings but it might
+help in some specific cases - notably on ppc64 (see bsc#1223982 and jsc#PED-8374).
+
+Accepted values are:
+
+- squashfs: Do not compress squashfs images in initrd.
+- modules: Do not compress kernel modules. Uncompress them if they are found compressed in kernel packages.
+- firmware: Do not compress kernel firmware. Uncompress them if they are found compressed in kernel packages.
--- a/gefrickel
+++ b/gefrickel
@@ -10,8 +10,13 @@ function err {
 }
 
 function squash {
+  local compression="-comp xz"
+  # see doc/compression.md
+  if [[ "$instsys_no_compression" =~ "squashfs" ]] ; then
+    compression="-no-compression"
+  fi
   mkdir -p parts
-  $mksquashfs $1 parts/$2 -comp xz -noappend -no-progress
+  $mksquashfs $1 parts/$2 $compression -noappend -no-progress
   chmod 644 parts/$2
 }
 
@@ -29,6 +34,11 @@ cd $dir
 
 if [ -L lib ]; then
   pfx="usr/"
+fi
+
+if [ -n "$instsys_no_compression" ] ; then
+  echo "$instsys_no_compression" > .no_compression
+  echo "using no compression in squashfs\n"
 fi
 
 # - - -  lib/{modules,firmware}  - - -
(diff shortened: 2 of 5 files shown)


lib/ReadConfig.pm
```diff
@@ -550,6 +550,41 @@ sub ReadRPM
...
+        SUSystem "find $dir -type l -name \*.$suffix -exec rename -sl .$suffix '' '{}' \\; -exec rename -l .$suffix '' '{}' \\;";
+      }
+
+      # unpack files
+      SUSystem "find $dir -type f -name \*.zst -exec zstd -d --quiet --rm '{}' \\;";
+      SUSystem "find $dir -type f -name \*.xz -exec xz -d '{}' \\;";
+      SUSystem "find $dir -type f -name \*.gz -exec gzip -d '{}' \\;";
+
+      my $broken = `find $dir -follow -type l`;
+      die "firmware uncompressing left broken symlinks:\n$broken\n" if $broken ne "";
```
hramrach, line 586: This will likely not work, the firmware already comes with a broken symlink, at least in the current installer images.

The broken case is when the number of broken symlinks increases after decompression.

lib/ReadConfig.pm
```diff
@@ -550,6 +550,41 @@ sub ReadRPM
...
+    # see doc/compression.md
+    if($ENV{instsys_no_compression} =~ /firmware/) {
+      my $dir = "$tdir/usr/lib/firmware";
+      $dir = "$tdir/lib/firmware" if ! -d $dir;
+
+      print "uncompressing kernel firmware...\n";
+
+      # rename symlinks
+      for my $suffix ("zst", "xz", "gz") {
+        SUSystem "find $dir -type l -name \*.$suffix -exec rename -sl .$suffix '' '{}' \\; -exec rename -l .$suffix '' '{}' \\;";
```
hramrach, line 577: I don't know what version of rename you are using. Mine does not document the -l option.

gefrickel
```diff
@@ -31,6 +36,11 @@ if [ -L lib ]; then
   pfx="usr/"
 fi
 
+if [ -n "$instsys_no_compression" ] ; then
+  echo "$instsys_no_compression" > .no_compression
+  echo "using no compression in squashfs\n"
```
hramrach, line 41: That's not what it actually does. In the extreme case the variable might be non-empty yet not match any of the options.

doc/compression.md
```diff
@@ -0,0 +1,13 @@
+# Adjusting compression
+
+The environment variable `instsys_no_compression` accepts a comma-separated
+list of values to omit compression in certain places.
```
hramrach, line 4: That's not how it's actually parsed but whatever. Passing what is documented should work.

lib/ReadConfig.pm
wfeldt: Not in TW. And I would see this as a packaging bug.

doc/compression.md
wfeldt: As above.

lib/ReadConfig.pm
wfeldt: It's there in SLE15 and TW.

hramrach: Not on Leap.

hramrach:
Also mutithreaded compression is not reproducible. Not when the number of threads still depends on number of CPUs, and not even when number of threads is fixed.

PR #727, opened 2024-07-22
make some compression settings configurable (bsc#1223982, jsc#PED-8374)

wfeldt:
## Task

Port https://github.com/openSUSE/installation-images/pull/726 to SLE15-SP6.

The difference here is that only the squashfs compression is disabled.

## Original task

- https://jira.suse.com/browse/PED-8374

Introduce environment variable `instsys_no_compression` to allow influencing compression strategy in some places. Notably for ppc64.

## Bonus

Limit compression threads in OBS to 4 threads, for reproducible builds.

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -775,6 +775,10 @@ export BUILD_DISTRIBUTION_NAME
 test ! -z "$BUILD_DISTRIBUTION_NAME"
 # build id (for linuxrc to start the correct instsys)
 export instsys_build_id=`bin/build_id`
+%ifarch ppc64 ppc64le
+# ppc64: optimize for smaller compressed initrd size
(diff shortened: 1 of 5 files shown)


obs/installation-images.spec
```diff
@@ -775,6 +775,10 @@ export BUILD_DISTRIBUTION_NAME
 test ! -z "$BUILD_DISTRIBUTION_NAME"
 # build id (for linuxrc to start the correct instsys)
 export instsys_build_id=`bin/build_id`
+%ifarch ppc64 ppc64le
+# ppc64: optimize for smaller compressed initrd size
+export instsys_no_compression=squashfs
```
hramrach, line 780: That does not look particularly useful

PR #730, opened 2024-07-31
Do not use non-deterministic multithreaded compression when optimizing compression (bsc#1223982 bsc#1223383 jsc#PED-8374).

hramrach:
Using multithreaded compression does not give deterministic results.

Not only can the file content vary but the size can also vary significantly.

Given there is need to fit the resulting image into certain limit using non-deterministic compression may result in some builds fitting and some not, at random, using same sources and tool.

### DIFF
--- a/lib/CompressImage.pm
+++ b/lib/CompressImage.pm
@@ -62,12 +62,15 @@ sub CompressImage
   print "compressing $image...\n";
 
   $prog_opt = '-cf9N' if $prog eq 'gzip';
+  $prog_opt = '-9 --check=crc32 -cf' if $prog eq 'xz';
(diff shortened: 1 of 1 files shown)


wfeldt:
installation-images builds are not reproducible builds anyway. Looking at a few build results I see sizes vary in the few ~100 byte range, if at all.
I don't see the point.

hramrach:
It will vary a lot more with the local --threads=0 build. I understand that for other platforms that don't mind the ramdisk size it's nice optimization but for ppc it will very likely break the images.

wfeldt:
In OBS it's limited to 4 threads which is what out workers have anyway.

hramrach:
Not for local builds.

PR #737, opened 2024-10-04
require libsolv-tools explicitly

hramrach:
previously it was implicitly installed via libsolv


wfeldt:
There must be a bsc reference; either in commit message or pr title.

wfeldt:
Cf. https://github.com/openSUSE/installation-images/pull/725

PR #740, opened 2024-10-22
add uio_hv_generic for hv_fcopy_uio_daemon

olafhering:

PR #752, opened 2025-01-16
add uio_hv_generic for hv_fcopy_uio_daemon (git-fixes)

olafhering:
Related to #740 and #742, for SLE15-SP7

### DIFF
--- a/etc/module.config
+++ b/etc/module.config
@@ -197,6 +197,7 @@ pmic_glink
 pmic_glink_altmode
 smp2p
 spmi-pmic-arb
+uio_hv_generic
 
 kernel/arch/.*/crypto/.*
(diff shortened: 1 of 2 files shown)


wfeldt:
There must be an explicit bsc/jira reference in at least one commit message or in the pr title.

olafhering:
Other changes went in with the `git-fixes` tag, I have adjusted the title.

wfeldt:
It does not help. A real bsc or jira tag is needed. Else the submission will be auto-rejected in the pipeline.

olafhering:
There is no number, but `git-fixes` should not be rejected.
There is no strict requirement to have this daemon running in the inst-sys. In case anyone really depends on that functionality, future quarterly updates can provide it again.

PR #762, opened 2025-06-24
etc: add mtu3 and onboard_usb_dev to module.config (bsc#1244269)

macpaul-lin-mtk:
mtu3 and onboard_usb_dev module was added to stable/Tumbleweed. Enumerate it in [other].


mbgg:
Looks good to me.

PR #781, opened 2026-02-02
Packaging: fix disabling mtime clamping

DimStar77:
The skelcd package is being extracted onto the FTP Servers and then
rsynced to the mirrors. rsync uses size/date information to identify
if a package needs to be synced. As RPM,for reproducible builds, sets
the mtime of the files to the date/time of the last changelog entry,
we break rsync on extracted files

This used not to be a problem, as OBS, when it extracted files out
of the VM, was not preserving mtimes. This had a negative effect on
republishing larger, yet stable trees like Leap 16.0. With this
fixed, it became apparent that the pre-existing attempt to disable
mtime clamping in the package no longer works

### DIFF
--- a/obs/installation-images.spec
+++ b/obs/installation-images.spec
@@ -25,7 +25,8 @@
 # Since they are all auto-generated files, so they keep on changing. Having the mtime
 # constant though causes issues for rsync - which uses this information by default to decide
 # wether there is something to sync or not
-%global clamp_mtime_to_source_date_epoch 0
+%global build_mtime_policy clamp_to_buildtime
+%global use_source_date_epoch_as_buildtime N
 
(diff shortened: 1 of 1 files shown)


DimStar77:
NOTE: in openSUSE:Factory, I have added this change as EmergencyFix over the weekend, see https://build.opensuse.org/package/rdiff/openSUSE:Factory/installation-images?linkrev=base&rev=243

The issue was detected by openQA https://openqa.opensuse.org/tests/5640197#step/welcome/4 after OBS changed the copy-from-vm mode in https://github.com/openSUSE/obs-build/commit/09e6b8c990719660e0f53c9f3df0bf3c5c59481f

PR #787, opened 2026-05-11
etc: add qnoc-qcs615 kernel module for Qualcomm QCS615/Talos Platform

fangwu1:
Add the qnoc-qcs615 interconnect kernel module for the Qualcomm QCS615/Talos platform to support SystemReady compliance. This allows generic Linux distribution ISO images to boot on QCS615-based hardware.

### DIFF
--- a/etc/module.config
+++ b/etc/module.config
@@ -207,6 +207,7 @@ spmi-mtk-pmif
 
 qnoc-sc7280
 qnoc-sa8775p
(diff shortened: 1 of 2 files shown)


fangwu1:
Hi @wfeldt, could you please take a look at this PR when you have a chance? It adds a kernel module for the Qualcomm QCS615 platform to support SystemReady compliance. Thanks!

